PR #48, opened 2020-06-08
Feature/sqs buffer

lukeindykiewicz:
PR moved from https://github.com/snowplow/snowplow/pull/4346
Check the previous desc ^

The sqs retry mechanism (af128bd) isn't the most beautiful solution, cause the messages are sent to sqs in batches. I could rewrite it in a more clean way, but I would have to send each message separately. I'm happy to rewrite that part if you have better ideas.


kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -228,8 +230,11 @@ class KinesisSink private (
     ()
   }
 
+  type BatchEntry = (ByteBuffer, String)
```
dilyand, line 233: Why not a class for easier field access?

lukeindykiewicz: I probably didn't want to change the original code from EventStorage too much, but I think this is good idea to use case class here! Thanks!

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -233,48 +288,148 @@ class KinesisSink private (
...
+    nextBackoff: Long
+  )(
+    retryErrorMsg: String
+  ): Unit =
+    maybeSqs match {
+      case Some(sqs) =>
+        log.info(
+          s"Sending ${failures.size} events from a batch to SQS buffer queue: ${sqs.sqsBufferName}"
+        )
+        putToSqs(sqs, failures)
```
benjben, line 340: Don't we want to add a callback with a logging of a potential error, just in case?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -233,48 +288,148 @@ class KinesisSink private (
...
+          s"${failures.size} failed events scheduled for retry (consider setting a SQS Buffer in config.hocon)"
+        )
+        scheduleBatch(failures, nextBackoff)
+    }
+
+  private def putToSqs(sqs: SqsClientAndName, batch: List[BatchEntry]): Future[Unit] =
+    Future {
+      log.info(s"Writing ${batch.size} messages to SQS queue: ${sqs.sqsBufferName}")
+      val sqsBatchEntries = batch.map(addUuids).map(toSqsBatchEntry)
+      val MaxSqsBatchSize = 10
```
benjben, line 354: Why 10 ?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -233,48 +288,148 @@ class KinesisSink private (
...
+      }
+    }
+
+  /**
+   *  UUIDs are generated for retry purpose. When batch fails, we want to retry (limited number of times)
+   *   only those entries that failed within the batch.
+   */
+  private val addUuids: (BatchEntry) => (BatchEntryWithId) = {
+    case BatchEntry(msg, key) => BatchEntryWithId(msg, key, java.util.UUID.randomUUID())
+  }
```
benjben, line 366: I remember that we already talked about this in another PR but needing a `case` to process the parameter of a function seems really weird to me.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -233,48 +288,148 @@ class KinesisSink private (
...
+    retry
+      .JitterBackoff(10, 1.second)
+      .apply { () =>
+        Future {
+          val batchRequest = createBatchRequest(sqs.sqsBufferName, current.values.toList)
+          val res = sqs.sqsClient.sendMessageBatch(batchRequest)
+          val failed = res.getFailed().asScala
+          if (failed.nonEmpty) {
+            val errors = failed.map(_.toString).mkString(", ")
+            log.error(s"Sending to SQS queue: ${sqs.sqsBufferName} failed with: $errors")
```
benjben, line 403:
```suggestion
            log.error(s"Sending to SQS queue [${sqs.sqsBufferName}] failed with errors [$errors]")
```
We usually use brackets to clearly indicate what comes from a variable

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -233,48 +288,148 @@ class KinesisSink private (
...
+  private def createBatchRequest(queueUrl: String, batch: List[SendMessageBatchRequestEntry]) =
+    new SendMessageBatchRequest()
+      .withQueueUrl(queueUrl)
+      .withEntries(batch.asJava)
+
+  private def sendToSqsWithRetry(
+    sqs: SqsClientAndName,
+    batchEntryGroup: List[(UUID, SendMessageBatchRequestEntry)]
+  ) = {
+    val current = TrieMap(batchEntryGroup: _*)
```
benjben, line 393: Given that SQS is to be used when there is a surged and collector is stressed, it would be nice to not need to copy all the data that we want to send each time.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
benjben:
```suggestion
  private val addUuids: (BatchEntry) => (BatchEntryWithId) = batchEntry => ...
```

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -14,59 +14,74 @@ package com.snowplowanalytics.snowplow.collectors.scalastream
...
 import com.amazonaws.services.kinesis.{AmazonKinesis, AmazonKinesisClientBuilder}
-
+import com.amazonaws.services.sqs.{AmazonSQS, AmazonSQSClientBuilder}
+import com.amazonaws.services.sqs.model.{
+  MessageAttributeValue,
+  SendMessageBatchRequest,
+  SendMessageBatchRequestEntry
+}
 import model._
+import java.util.UUID
```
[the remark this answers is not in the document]
lukeindykiewicz: ok, I'll try to remember that.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -14,59 +14,74 @@ package com.snowplowanalytics.snowplow.collectors.scalastream
...
   def createAndInitialize(
     kinesisConfig: Kinesis,
     bufferConfig: BufferConfig,
     streamName: String,
+    sqsBufferName: Option[String],
     executorService: ScheduledExecutorService
   ): Either[Throwable, KinesisSink] = {
-    val client = for {
+    val clients = for {
       provider <- getProvider(kinesisConfig.aws)
```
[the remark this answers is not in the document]
lukeindykiewicz: It's not the only change needed, the same val is used few lines below. Renamed.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -182,8 +230,15 @@ class KinesisSink private (
     ()
   }
 
+  case class BatchEntry(msg: ByteBuffer, key: String)
```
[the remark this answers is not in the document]
lukeindykiewicz: I followed the aws api convntion here, as we pass it to methods from the api. This is how it is named there. It's not an sqs event, it's rather SnowplowEvent, so `Event` in short.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -196,7 +251,7 @@ class KinesisSink private (
         )
```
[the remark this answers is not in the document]
lukeindykiewicz: It doesn't happen here at all. I think this should be fixed in other issue/PR.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -182,8 +230,15 @@ class KinesisSink private (
     ()
   }
 
+  case class BatchEntry(msg: ByteBuffer, key: String)
+  case class BatchEntryWithId(
+    msg: ByteBuffer,
+    kwy: String,
```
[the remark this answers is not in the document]
lukeindykiewicz: I don't understand what do you mean? It's basically the case class instead of repeating the types.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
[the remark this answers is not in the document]
lukeindykiewicz: https://github.com/snowplow/stream-collector/issues/54

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: Great catch! It's weird, cause it was a type alias before and I missed this place needs more love after changing from type alias to case class. Thanks.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: Really? Sounds weird, brackets indicates tables usually.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: It's max size of batch for sqs

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -233,48 +288,148 @@ class KinesisSink private (
...
         }
       }
     }
 
-  private def multiPut(name: String, batch: List[(ByteBuffer, String)]): Future[PutRecordsResult] =
+  private def sendToSqsOrRetryToKinesis(
+    failures: List[BatchEntry],
+    nextBackoff: Long
+  )(
+    retryErrorMsg: String
```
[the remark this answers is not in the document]
lukeindykiewicz: My reasoning was that the error message is completely different from what's actually important in this function.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: It's done in retry mechanism in `sendToSqsWithRetry`, so further in function calls.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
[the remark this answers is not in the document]
benjben: Exactly. My point is that `EventWithId` should just be `Event` with `Id` so that if you update the fields of `Event` it needs to happen only in one place

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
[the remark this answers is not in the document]
benjben: But programmatic-wise there is no difference (contrary to implicit parameters or higher order functions for instance)

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
benjben: Would be nice to add a comment that says it

lukeindykiewicz: but the name says exactly that, isn't it?

lukeindykiewicz:
> I think that we can not release without #54 , or the collector becomes lossy.
> Please check with Product.

@benjben  How is this more lossy than it is right now? I'm adding the surge protection feature, so there is one more step to not loose events at the moment, so why do you say it's going to be more lossy?
I'm saying it's more/less lossy, not lossy/non-lossy, cause the collector is lossy at the moment. This feature prevents some craches (that were loosing events), so I would say it's less lossy than it was before. 
I'm not sure what would you like to discuss with Product team? I'm of course happy to discuss, just point to what this discussion should be about?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: It would be nice, that's true. Tests are showing that cpu utilization is less than it was before. Of course we would see the real numbers on production. The solution here is not the prettiest, but it lets us send batches of events instead of sending one event per api call. We might improve it in the future though. Unless you have some good ideas how to make it readable and not copy the list?

benjben:
@lukeindykiewicz at the moment if an event is too big to be inserted into Kinesis (1MB I think), a bad row is created and emitted for this event, so we know that it was sent. As I understand, if an event is too big for SQS, we just discard it and we don't know if the collector received or not. Did I misunderstand something?

lukeindykiewicz:
@benjben I don't think so. Check https://github.com/snowplow/stream-collector/blob/363fb4397805f10823dc223347dc4fc61ef74afc/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala#L193
Looks like no bad row event is generated, only simple log message. I'm not changing this.

benjben:
@lukeindykiewicz If an event is too big for Kinesis, a `SizeViolation` bad row is created, look:

https://github.com/snowplow/stream-collector/blob/363fb4397805f10823dc223347dc4fc61ef74afc/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/utils/SplitBatch.scala#L134

In SQS documentation :

> Message payloads can contain up to 256KB of text in any format

If a message of 512KB can't be inserted into Kinesis, SQS will log a message and just discard it. It would be really great if we could get a `BadRow` when this happens.

lukeindykiewicz:
@benjben Thanks for pointing to this split method, which is way before the code I was changing. I think the solution would be just to change the `MaxBytes` if the sqs buffer is enabled. From what TechOps checked for last 6 months for client the messages are mostly small, not reaching the 256k limit.

benjben:
Makes sense to change `MaxBytes` if SQS buffer is enabled. Just one thing, we need to take into account base64 encoding, hopefully messages stay < 256k with the encoding.

lukeindykiewicz:
The same question remains for Kinesis, so I assume it's done correctly already. I will not change this behavior.

benjben:
I think that we don't base64 encode the messages that we send to Kinesis

lukeindykiewicz:
That's true - you're right. I'll check that. Thanks!

lukeindykiewicz:
The `pending` info is an error, click `Details` to see the build is green.

chuwy:
Unless I'm missing something crucial I agree with @lukeindykiewicz - we should not add one more ticket on top of what he's doing now. Collector is not becoming more lossy because of this change and it can be pushed back to the next release.

benjben:
@chuwy 

> Collector is not becoming more lossy because of this change

It was but Luke fixed that in https://github.com/snowplow/stream-collector/pull/48/commits/a256e58054324629664e7d3f145bfff740ee61e4#diff-67fdfd8fe30fa650f6a58346af2157feR186

PR #53, opened 2020-06-10
Add snyk monitor (closes #52)

### LINKED ISSUE #52: Add snyk monitor

lukeindykiewicz:
This adds snyk monitor as gh action, just because it's a lot easier than configure the Travis at the moment and it should happen only on push to master I believe, so it's not needed to install snyk via npm in travis (and waste time for it on every build).

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -1,0 +1,20 @@
+name: Snyk
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  security:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      
+    - name: Run Snyk to check for vulnerabilities
+      uses: snyk/actions/scala@master
+      with:
+        command: monitor
+        args: --project-name=stream-collector
+      env:
+        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}


paulboocock:
This needs updating slightly as I've noticed it doesn't pick up the project name automatically, but uses the folder name. You can fix this however:

```
    - name: Run Snyk to check for vulnerabilities
      uses: snyk/actions/scala@master
      with:
        command: monitor
        args: --project-name=stream-collector
```

Additionally, Dilyan noticed some issues with sbt 1.3 in his testing. Not sure if you had the same issue when testing this locally?
https://snplow.atlassian.net/wiki/spaces/EN/pages/1205600277/Adding+Snyk+to+a+Project#sbt-1.3%2B-incompatibility

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,19 @@
...
+  security:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      
+    - name: Run Snyk to check for vulnerabilities
+      uses: snyk/actions/scala@master
+      with:
+        command: monitor
```
benjben, line 17:
```suggestion
        command: monitor
        args: --project-name=stream-collector --org=data-cap
```

paulboocock: You don't need to specify --org=data-cap if the service account token is generated against the Data Cap organisation in Snyk

benjben: Good, to know! Thanks @paulboocock!

chuwy, line 17: Mine works without `monitor`. Do we need it? https://github.com/snowplow/enrich/blob/release/1.3.0/.github/workflows/snyk.yml

chuwy: But somehow you do need to specify `--project-name` if I'm not mistaken.

paulboocock: Yes, it defaults to `test` which doesn't upload the report to the Snyk UI. Currently, we're looking at running `monitor` so we have full visibility of our projects in the Snyk UI. 

`test` is useful if you want your build to fail if there are any vulnerabilities, which we will likely add to our build processes in the future once we're more comfortable with where we are at with Snyk.

paulboocock: You'll probably need `args: --project-name=stream-collector` too. It tries to find the project name from config, or it defaults to the containing folder name. Unfortunately Github actions maps this into a `workspace` folder in the docker image that's running behind the scenes. So specifying `--project-name` just guarentees we get nice names in the Snyk UI.

lukeindykiewicz:
> Not sure if you had the same issue when testing this locally?

yeah, but I added the global plugin and it worked. Now I checked the version, I use the same as Dilyan mentioned.

lukeindykiewicz:
> As Paul mentioned there might be problems with SBT 1.3, but it worked when I added latest RC of sbt-dependency-graph to project plugins.

Thanks Anton! I already had that :)

.github/workflows/snyk.yml
lukeindykiewicz: @benjben I think this is explained, could you change PR status, please?

PR #65, opened 2020-08-06
Disable default redirect (close #64)

### LINKED ISSUE #64: Disable default redirect

lukeindykiewicz:
PR related to https://snplow.atlassian.net/browse/PE-731

TF changes here: https://github.com/snowplow-devops/terraform-modules/pull/1412

### DIFF
--- a/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
+++ b/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
@@ -156,7 +156,7 @@ package model {
     cors: CORSConfig,
     streams: StreamsConfig,
     prometheusMetrics: PrometheusMetricsConfig,
-    enableDefaultRedirect: Boolean = true,
+    enableDefaultRedirect: Boolean = false,
     ssl: SSLConfig = SSLConfig()
   ) {
     val cookieConfig = if (cookie.enabled) Some(cookie) else None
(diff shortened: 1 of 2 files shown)


lukeindykiewicz:
@dilyand 

The `enableDefaultRedirect` config param is tested in here:
https://github.com/snowplow/stream-collector/blob/237c50e933cfcc8b56e2a392abf20404622a821a/core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRouteSpec.scala#L106

The config can't have `None` I think, cause it's modeled like this:
https://github.com/snowplow/stream-collector/blob/52d715d1e6f9b23e50c2e88973a4414a4a253894/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala#L159

The parsing is done in:
https://github.com/snowplow/stream-collector/blob/52d715d1e6f9b23e50c2e88973a4414a4a253894/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala#L67

We have default value in `enableDefaultRedirect` in `CollectorConfig`, so if there is no value specified in config file, the default value will be used.

dilyand:
@lukeindykiewicz 

We don't have `enableDefaultRedirect` set at all in: https://github.com/snowplow/stream-collector/blob/luke/disable_default_redirect/core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/TestUtils.scala

And we don't use this config in the `CollectorRouteSpec` test, where we explicitly set it to `false`. I thought it might be a good idea to test it a config that simply omits the parameter, but it's probably overkill.

lukeindykiewicz:
> We don't have enableDefaultRedirect set at all in

Yeah, exactly what I meant - it's not explicitly set, which means it uses the default value.


> And we don't use this config in the CollectorRouteSpec test, where we explicitly set it to false

Yeah, that's why I said it's tested there, cause it's set explicitly.

I mentioned two things, which weren't separated good enough - sorry if that wasn't clear.
I should have start with: thanks for review @dilyand!

PR #66, opened 2020-08-19
Release 2.0.0 to MASTER

lukeindykiewicz:


.github/workflows/snyk.yml
```diff
@@ -0,0 +1,20 @@
+name: Snyk
```
dilyand, line 1: Have you tested this @lukeindykiewicz ? I think it won't work without adding the `sbt-dependency-graph` plugin to the project.

To avoid adding one more plugin, you can forego the default Snyk action and instead do it like here: https://github.com/snowplow/scala-maxmind-iplookups/blob/master/.github/workflows/snyk.yml.

build.sbt
```diff
@@ -46,7 +46,7 @@ lazy val commonDependencies = Seq(
 lazy val buildSettings = Seq(
   organization  :=  "com.snowplowanalytics",
   name          :=  "snowplow-stream-collector",
-  version       :=  "1.0.1",
+  version       :=  "2.0.0",
```
dilyand, line 49: I think this change should be part of the `Prepare for release` commit. We used to have it as a separate step in the umbrella releases, because it was possible to release a new `R11x` version of `snowplow` and *not* bump the collector. But now that it's in its own repo, we should follow the practice from all our others single-project repos.

CHANGELOG
```diff
@@ -1,3 +1,13 @@
+Release 2.0.0 (2020-08-19)
```
dilyand, line 1: Needs updating.

CHANGELOG
```diff
@@ -1,3 +1,13 @@
...
+--------------------------
+Bump to 2.0.0 (close #61)
```
dilyand, line 3: Should not be a separate commit.

dilyand (changes requested):
Hey @lukeindykiewicz , looks good. I've left a couple of comments. Nothing major but I think they need to be addressed before releasing.

I remember reviewing the sqs buffer changes in #48 and the redirect path changes in #65, and I have assumed that this PR is identical to what I have already green-ticked. But let me know if you have made any changes in the interim that need closer attention.

build.sbt
benjben: Agreed with Dilyan.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -233,52 +288,150 @@ class KinesisSink private (
...
+  )(
+    retryErrorMsg: String
+  ): Unit =
+    maybeSqs match {
+      case Some(sqs) =>
+        log.info(
+          s"Sending ${failures.size} events from a batch to SQS buffer queue: ${sqs.sqsBufferName}"
+        )
+        putToSqs(sqs, failures)
+        ()
```
benjben, line 341: Don't we want to wait for the result of `putToSqs` instead of discarding the `Future`? What if all the writes to SQS fail?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -233,52 +288,150 @@ class KinesisSink private (
...
+      case None =>
+        log.error(retryErrorMsg)
+        log.warn(
+          s"${failures.size} failed events scheduled for retry (consider setting a SQS Buffer in config.hocon)"
+        )
+        scheduleBatch(failures, nextBackoff)
+    }
+
+  private def putToSqs(sqs: SqsClientAndName, batch: List[Event]): Future[Unit] =
+    Future {
```
benjben, line 351: Why wrapping it in a `Future` ?

build.sbt
lukeindykiewicz: done

.travis/deploy.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -ex
+
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+
+sbt "project kinesis" docker:publish && sbt "project pubsub" docker:publish && sbt "project kafka" docker:publish && sbt "project nsq" docker:publish && sbt "project stdout" docker:publish
```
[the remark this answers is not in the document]
lukeindykiewicz: Just a convention I think, all repos has it in deploy.sh

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: The batch call to SQS is retried up to 10 times. This is already the backup, so unless we want to implement some storage backup for the backup we are pretty much out of options. We don't want to keep it in memory when sqs buffer is enabled, as this is exactly the problem we're fixing with the whole sqs buffer.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: To be fast, we don't have to wait for the result.

.github/workflows/snyk.yml
lukeindykiewicz: Thanks @dilyand for spotting this!

lukeindykiewicz: fixed in 314c3de

lukeindykiewicz:
I added two commits that adds the plugin needed by snyk and also yet another bump of jackson was needed, as the bumped version was already vulnerable again.
I've added (to be removed before merging) rc version to check if everything works.

dilyand (changes requested):
I think this needs three finishing touches:

1.) These two commits (and their corresponding issues) should be merged:
- Add snyk monitor (close #52)
- Add sbt-dependency-graph plugin for snyk (close #69)

The only reason we're adding the plugin is so that snyk monitor will work. It doesn't make sense to have them as separate items.

2.) These two should also be merged (and their corresponding issues):
- Bump vulnerable libs (close #56)
- Bump jackson library due to yet another vulnerability (close #70)

Issues should represent changes since the last published version, but #70 is a change since #56 which is not part of the latest published version.

3.) The CHANGELOG should be updated:
- remove 'Bump to 2.0.0 (close #61)', which is now part of the 'Prepare for release'  commit
- reflect the final state of commits after the merges in 1 and 2
- a milestone should be created with the issues we are closing in this release
- redundant issues (like 70, 69, 61) should be removed.

To make it easier to track progress:

- [x] merge snyk commits
- [x] merge lib bump commits
- [ ] milestone 2.0.0
- [ ] CHANGELOG cleanup
- [ ] final check of 1:1:1 sync between issue tracker, CHANGELOG and commit history

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
benjben: > The batch call to SQS is retried up to 10 times..
If it fails after 10 times, we must have a log message to indicate that we are dropping events.

benjben (changes requested):
We are missing a log message to indicate that we are dropping events after 10 unsuccessful attempt to write to SQS

lukeindykiewicz:
@benjben The log messages shows what failed and what succeeded. Nothing changes since this PR was merged: https://github.com/snowplow/stream-collector/pull/48
Around this place: https://github.com/snowplow/stream-collector/pull/66/files#diff-67fdfd8fe30fa650f6a58346af2157feR402

benjben:
@lukeindykiewicz still we don't know when the collector is dropping messages.

dilyand:
> @lukeindykiewicz still we don't know when the collector is dropping messages.

Ben is right, we can potentially see 9 log messages for SQS failures but the write can succeed on the 10th attempt. In order to know for sure if we're dropping events, we'll need to count the failures for each batch. It would be hard to build reliable monitoring and alerting on top of that.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -138,10 +175,15 @@ class KinesisSink private (
...
+  maybeSqs: Option[SqsClientAndName]
 ) extends Sink {
-  // Records must not exceed MaxBytes - 1MB
-  val MaxBytes = 1000000
+  // Records must not exceed MaxBytes - 1MB (for Kinesis)
+  // When SQS buffer is enabled MaxBytes has to be 256k,
+  // but we encode the message with Base64 for SQS, so the limit drops to 192k
+  val SqsLimit = 192000 // 256000 / 4 * 3
+  val KinesisLimit = 1000000
+  override val MaxBytes = if (maybeSqs.isDefined) SqsLimit else KinesisLimit
```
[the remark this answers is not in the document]
dilyand: There is a big difference between the MaxBytes allowed by Kinesis and SQS, which means there's a tradeoff on using surge protection -- your events must be way leaner. Not something to address in this PR but we should make it very clear in all documentation and comms (with customers / users but also Support / Tech Ops / Product).

dilyand (approved):
LGTM!

If possible, we should test this with a new rc to ensure nothing broke but also to see if we can make it drop events and log that.

Also, just as a point of process, the issues in the 2.0.0 milestone must stay open. They are auto-closed when the release is merged to `master` (that is the point of the `close #XX` in the commit messages). This is documented here: https://github.com/snowplow-incubator/engineering-resources/blob/master/release-process.md.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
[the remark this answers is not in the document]
lukeindykiewicz: That is reflected in announcement and blog post. Thanks for good point.

lukeindykiewicz:
> If possible, we should test this with a new rc to ensure nothing broke but also to see if we can make it drop events and log that.

definitely

> Also, just as a point of process, the issues in the 2.0.0 milestone must stay open. They are auto-closed when the release is merged to master (that is the point of the close #XX in the commit messages). This is documented here: https://github.com/snowplow-incubator/engineering-resources/blob/master/release-process.md.

The issues were closed when the feature PR was merged into develop from feature branch I believe. The gitflow workflow is not 100% compatible with what is described in the release-process doc.

dilyand:
> The issues were closed when the feature PR was merged into develop from feature branch I believe. 

Should we change the default branch then? Up until a release takes place, there is a possibility to discover bugs that require the issue to be reopened.

benjben:
We should definitely change the default branch to `master` ! @chuwy do you have the power to do it ?

lukeindykiewicz:
I have the admin rights here, so I can change it.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -233,52 +288,160 @@ class KinesisSink private (
...
+  private def putToSqs(sqs: SqsClientAndName, batch: List[Event]): Future[Unit] =
+    Future {
+      log.info(s"Writing ${batch.size} messages to SQS queue: ${sqs.sqsBufferName}")
+      val sqsBatchEntries = batch.map(addUuids).map(toSqsBatchEntry)
+      val MaxSqsBatchSize = 10
+      val MaxRetries = 10
+      sqsBatchEntries.grouped(MaxSqsBatchSize).foreach { batchEntryGroup =>
+        sendToSqsWithRetry(sqs, batchEntryGroup, MaxRetries).transform {
+          case failure @ Failure(ex) =>
+            log.info(s"Sending to sqs failed with exception: $ex")
```
benjben, line 359: In this case, are events also dropped? If yes it would be great to include it in the log line.

benjben (approved):
:+1:

PR #67, opened 2020-08-19
Release 2.0.0 back to develop

lukeindykiewicz:


dilyand:
@lukeindykiewicz I think this pull request should be closed. Once the release is out and all changes merged to `master`, then you can open a PR from the release branch to `develop`. But having two PRs open and in review simultaneously is confusing and also leads to double work, since any changes in one you must also ensure are also in the other.

benjben:
I agree with @dilyand. I would even say that once we have merged to `master`, we can manually merge to `develop`.

dilyand:
For sure. Once the release is out, the merge back to `develop` is a formality and does not need to go through review.

lukeindykiewicz:
I don't agree, because this release to master should also be just a formality, so both open PRs would be open just for a short while. 

It would not confuse anyone than. It is just a special situation as it was first left untouched for two weeks, than confused of what we do here and recently we found some issues, so it's still open.

I can close this particular one if that will make you happy, but as a general rule I don't agree.

PR #76, opened 2020-10-29
Increase number of Kinesis put retries when surge protection is on

lukeindykiewicz:


kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -350,82 +394,61 @@ class KinesisSink private (
...
-              new MessageAttributeValue()
-                .withDataType("String")
-                .withStringValue(eventWithId.event.key)
-          ).asJava
-        )
-        .withId(eventWithId.id.toString)
-      (eventWithId.id, batchReqEntry)
+  private val toSqsBatchEntry: Event => SendMessageBatchRequestEntry = { event =>
+    val b64EncodedMsg = encode(event.msg)
+    val batchReqEntry = new SendMessageBatchRequestEntry(UUID.randomUUID.toString, b64EncodedMsg)
```
dilyand, line 413: Do we still need the UUID if we're not doing 'manual' retries?

dilyand (commented):
Hey @lukeindykiewicz , LGTM! I have some questions but I'm not sure if they are things that need to be addressed now or not. Have a look and let me know what you think.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -150,6 +177,24 @@ object KinesisSink {
...
+        .withClientConfiguration(
+          new ClientConfiguration().withRetryPolicy(
+            new RetryPolicy(
+              new RetryCondition {
+                override def shouldRetry(
+                  originalRequest: AmazonWebServiceRequest,
+                  exception: AmazonClientException,
+                  retriesAttempted: Int
+                ): Boolean =
+                  retriesAttempted < 10
```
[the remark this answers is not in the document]
lukeindykiewicz: > If I'm not missing anything, it means that if something cannot be inserted in SQS 10 times, we'll drop it?

Yes, we will stop retrying after 10 times, so that means drop it.

So currently in master (2.0.0) we also have 10 retries with JitterBackoff, set up with 1 second delay. 
```scala
 retry.JitterBackoff(maxRetries, 1.second)
```
The 10 retries is backed in how we determine when to stop retrying. 

In `2.0.1` we have 
```scala
new PredefinedBackoffStrategies.FullJitterBackoffStrategy(1000, 5 * 3600),
```
which means we start with 1 sec, but max backoff is limited to 5 mins (arbitrary numbers).

We have to balance somehow for how long we retry to not keep too much in memory to not kill collector I think.

> Do we expect that this will happen only when SQS is unavailable? If yes, do 10 attempts give it reasonable time to recover?

This is only a solution for AWS APIs glitches, so only when it's unavailable for short period of time. It will not help us when there is a sqs outage.

> Are there any other reasons we can foresee for a record not being able to be inserted in SQS? Does it make sense to generate a bad row here?

We validate the event earlier if it's not too big - that's one thing that comes to my mind. 
We also check if the sqs queue exists.
I don't know if there are other reasons?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -302,7 +346,7 @@ class KinesisSink private (
       multiPut(streamName, batch).onComplete {
         case Success(s) => {
           val results = s.getRecords.asScala.toList
-          val failurePairs = batch zip results filter { _._2.getErrorMessage != null }
+          val failurePairs = batch.zip(results).filter(_._2.getErrorMessage != null)
```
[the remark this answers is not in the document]
lukeindykiewicz: Yes, the unlimited number of retries is because we keep failed inserts to Kinesis in memory to retry again later. So, we have two levels of retries now. One is baked in Kinesis client, the other is the old mechanism that we always had.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -350,82 +394,61 @@ class KinesisSink private (
...
+    val batchReqEntry = new SendMessageBatchRequestEntry(UUID.randomUUID.toString, b64EncodedMsg)
+      .withMessageAttributes(
+        Map(
+          "kinesisKey" ->
+            new MessageAttributeValue()
+              .withDataType("String")
+              .withStringValue(event.key)
+        ).asJava
+      )
+    batchReqEntry
```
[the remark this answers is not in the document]
lukeindykiewicz: :+1:

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: We don't need this UUID, but there are only two constructors for `SendMessageBatchRequestEntry`, one is parameterless, the other has both id and body, so we set some id and don't use it afterwards.

dilyand: Cool, maybe worth noting that in a comment?

dilyand (approved):
👍🏻

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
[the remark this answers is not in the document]
benjben: > We have to balance somehow for how long we retry to not keep too much in memory to not kill collector I think.

It seems to me that the number of minutes (5) and the maximum numbers of retries (10) should be able to set via configuration. If someone is very worried about losing events, he/she might want to give the collector a lot of memory so that it can keep events in-memory for a longer time in case of Kinesis and SQS outage.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -127,6 +131,29 @@ object KinesisSink {
...
+                  exception: AmazonClientException,
+                  retriesAttempted: Int
+                ): Boolean =
+                  retriesAttempted < 10 &&
+                    (exception match {
+                      case _: ProvisionedThroughputExceededException => false
+                      case _ => true
+                    })
+              },
+              new PredefinedBackoffStrategies.FullJitterBackoffStrategy(1000, 5 * 3600),
```
benjben, line 150: See https://github.com/snowplow/stream-collector/pull/76#discussion_r525459144

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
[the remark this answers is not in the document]
benjben: So if SQS buffer is activated, the first level of retries (in Kinesis client) still occurs, but not the second, right?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -350,82 +394,60 @@ class KinesisSink private (
...
-          case s @ Success(Some(_)) => s
+      batch
+        .map(toSqsBatchEntry)
+        .grouped(MaxSqsBatchSize)
+        .foreach { batchEntryGroup =>
+          sendToSqs(sqs, batchEntryGroup).transform {
+            case failure @ Failure(ex) =>
+              log.info(s"Sending to sqs failed with exception: $ex")
+              failure
+            case s @ Success(_) => s
```
benjben, line 406: Why do we have return values if `putToSqs(sqs: SqsClientAndName, batch: List[Event])` returns `Future[Unit]` ?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -235,7 +280,6 @@ class KinesisSink private (
   }
 
   case class Event(msg: ByteBuffer, key: String)
```
benjben, line 282: Would be great to say that this `key` is for Kinesis.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -350,82 +394,60 @@ class KinesisSink private (
...
-      }
-  }
+    batchEntryGroup: List[SendMessageBatchRequestEntry]
+  ) =
+    Future {
+      val batchRequest = createBatchRequest(sqs.sqsBufferName, batchEntryGroup)
+      val res = sqs.sqsClient.sendMessageBatch(batchRequest)
+      val failed = res.getFailed().asScala
+      if (failed.nonEmpty) {
+        // It could be improved even more by storing events (that failed to be sent to SQS) to some persistance storage
```
benjben, line 438: Great idea !

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -350,82 +394,60 @@ class KinesisSink private (
...
+        val errors = failed.map(_.toString).mkString(", ")
+        log.error(
+          s"Sending to SQS queue [${sqs.sqsBufferName}] failed with errors [$errors]. Dropping events."
+        )
+        log.info(
```
benjben, line 443: Why don't we put this in the above log ?

benjben (commented):
As I understand we have this PR because Kinesis can sometimes be unavailable for a few sec and in this case we want to retry the writing? Please do not hesitate to give some context when you create an issue.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: That's different logging levels. I thought the warning should be just warning. Could be merged if that is not the case, of course.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: The cases are inside `.transform` function. The `Unit` comes from `foreach` in 401.

lukeindykiewicz:
> So if SQS buffer is activated, the first level of retries (in Kinesis client) still occurs, but not the second, right?

yes, exactly as you said

PR #96, opened 2020-11-26
remove stream existance check

lukeindykiewicz:


kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -138,36 +136,17 @@ object KinesisSink {
                 exception: AmazonClientException,
                 retriesAttempted: Int
               ): Boolean =
-                retriesAttempted < 10 &&
-                  (exception match {
-                    case _: ProvisionedThroughputExceededException => false
-                    case _                                         => true
-                  })
+                false
```
dilyand, line 139: Is the assumption here that we don't need to check if this should be retried because we would only deploy this asset when Kinesis is down and so we know that there is no point in retrying?

build.sbt
```diff
@@ -45,7 +45,7 @@ lazy val commonDependencies = Seq(
 lazy val buildSettings = Seq(
   organization := "com.snowplowanalytics",
   name := "snowplow-stream-collector",
-  version := "2.0.1",
+  version := "2.0.2-rc1",
```
dilyand, line 48: Maybe the tag can make it clearer that this is not part of the normal release process and this will not be merged to `master` but just used from the branch? On that note, maybe it shouldn't be on a `fix/*` branch but rather some sort of `emergency/kinesis-down` branch?

dilyand (approved):
This looks alright to me, though I think we should make it clearer that it will never be promoted to 2.0.2 and will remain an `rc` asset for emergencies only.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
lukeindykiewicz: Yes, exactly as you said. Don't loose time on retries, when we know kinesis is down.

build.sbt
lukeindykiewicz: Totally agree, would be good to have such convention. I should use hotfix at least, as in gitflow workflow, sorry.

dilyand:
The underlying issue was solved in https://github.com/snowplow/stream-collector/issues/100.

PR #101, opened 2020-12-04
SQS existence check

dilyand:


kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -56,9 +56,16 @@ object KinesisSink {
     val clients = for {
       credentials <- getProvider(kinesisConfig.aws)
       kinesisClient = createKinesisClient(credentials, kinesisConfig.endpoint, kinesisConfig.region)
-      _ <- if (streamExists(kinesisClient, streamName)) true.asRight
-      else new IllegalArgumentException(s"Kinesis stream $streamName doesn't exist").asLeft
       sqsClientAndName <- sqsBuffer(sqsBufferName, credentials, kinesisConfig.region)
+      _ <- if (streamExists(kinesisClient, streamName)) {
+        true.asRight
+      } else if (sqsClientAndName.isDefined) {
```
chuwy, line 62: Shouldn't we also check if it really exists and not just defined?

dilyand: Good point, I've added that check.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -206,6 +208,36 @@ object KinesisSink {
...
+    } else if (sqs.isDefined && sqsQueueExists(
+                 sqs.get.sqsClient,
+                 sqs.get.sqsBufferName
+               )) {
+      true.asRight
+    } else {
+      new IllegalArgumentException(
+        s"Kinesis stream $streamName doesn't exist and no SQS buffer is configured. At least one of the two is required. If you have configured an SQS buffer, double-check if the queues specified in your config file actually exist."
+      ).asLeft
+    }
```
chuwy, line 240:
```suggestion
private def runChecks(kinesisClient: AmazonKinesis, streamName: String, sqs: Option[SqsClientAndName]): Either[IllegalArgumentException, Unit] =
  sqs match {
    case Some(clientAndName) => 
      if (sqsQueueExists(clientAndName.sqsClient, clientAndName.sqsBufferName)) 
        ().asRight
      else new IllegalArgumentException(s"SQS queue ${clientAndName.sqsBufferName} is defined, but not exists").asLeft  // Assuming this is also invalid state
    case None if streamExists(kinesisClient, streamName) =>
      ().asRight
    case None =>
      new IllegalArgumentException(
        s"Kinesis stream $streamName doesn't exist and no SQS buffer is not configured. At least one of the two is required. If you have configured an SQS buffer, double-check if the queues specified in your config file actually exist."
      ).asLeft
  }
```

1. `Boolean` doesn't bear any information here
2. I think `Option.get` should not be used if there's even slight chance to avoid it
3. The orginal error message might be misleading because in final `else` we don't know if SQS buffer is configured, but not exists or not configured

dilyand: > I think Option.get should not be used if there's even slight chance to avoid it

You purist.

I've taken this onboard.

chuwy (approved):
👍

dilyand:
We now `WARN` rather than throw.

dilyand:
We now `WARN` rather than throw.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -206,6 +209,38 @@ object KinesisSink {
...
+    * Check whether an SQS queue (buffer) exists
+    * @param name Name of the queue
+    * @return Whether the queue exists
+    */
+  private def sqsQueueExists(client: AmazonSQS, name: String): Boolean =
+    try {
+      client.getQueueUrl(name)
+      true
+    } catch {
+      case _: QueueDoesNotExistException => false
```
benjben, line 222: There are no other types of exceptions to handle, e.g. timeout or network issues? I wonder if we should add a `case NonFatal(e)`

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -206,6 +209,38 @@ object KinesisSink {
...
+    streamName: String,
+    sqs: Option[SqsClientAndName]
+  ): Unit = {
+    lazy val log = LoggerFactory.getLogger(getClass())
+
+    // format: off
+    sqs match {
+      case Some(clientAndName) =>
+        if (sqsQueueExists(clientAndName.sqsClient, clientAndName.sqsBufferName)) ()
+        else log.warn(s"SQS queue ${clientAndName.sqsBufferName} is defined in config file, but does not exist.")
```
benjben, line 236: Even if SQS queue is defined, I think that we want to check the existence of Kinesis stream and log a warning if it doesn't exist.

benjben (changes requested):
I think that we should rename the issue to something like `Do not crash and log a warning if Kinesis stream and SQS queue don't exist`

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
dilyand: Good point, I've added that log message.

benjben (changes requested):
In the scenario where collector starts during an AWS outage, so collector starts and logs that Kinesis and SQS aren't available. Then it still receives messages and keeps trying to insert them into Kinesis/SQS and when they will become available it will automatically empty its buffer, right?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -206,6 +209,40 @@ object KinesisSink {
...
+    if (!kExists) log.error(s"Kinesis stream $streamName doesn't exist or isn't available.")
+
+    // format: off
+    sqs match {
+      case Some(clientAndName) =>
+        if (sqsQueueExists(clientAndName.sqsClient, clientAndName.sqsBufferName)) ()
+        else log.error(s"SQS queue ${clientAndName.sqsBufferName} is defined in config file, but does not exist or isn't available.")
+      case None =>
+        if (kExists) ()
+        else log.error(s"SQS buffer is not configured.")
```
benjben, line 242: I think that it should be a `warn` rather than `error`, because it can be expected behavior that SQS is not used (it will mostly be the case)

benjben (approved):
:+1:

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
dilyand: Ah, but we only error if there's also no Kinesis.

PR #118, opened 2021-02-16
Release/2.1.2

### LINKED ISSUE #117: Prevent network_userid being captured when SP-Anonymous header is set
Currently, if a `nuid` cookie is sent on a request from the browser the collector will use it as the `network_userid` regardless of whether the `SP-Anonymous` header is present. 

This means for users who have previously visited a site and accepted cookies but then opted out we still capture the `network_userid` because the cookie is present in the browser (and we can't delete it in JavaScript as it is `HttpOnly`).

We should only capture the `network_userid` when the `SP-Anonymous` header is not sent, any presence of the `SP-Anonymous` header should prevent the `network_userid` from being captured.

paulboocock:
This PR resolves an anonymous tracking issue which allows the network_userid to still be captured when a previously existing cookie is present in the browser. Closes #117.


core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -243,7 +243,7 @@ class CollectorService(
     userAgent.foreach(e.userAgent   = _)
     refererUri.foreach(e.refererUri = _)
     e.hostname      = hostname
-    e.networkUserId = networkUserId
+    e.networkUserId = anonymize(networkUserId, spAnonymous).orEmpty
```
chuwy, line 246: Given that it's the only place where the function is used - I think we can just return empty string?

Or even abandon the function altogether:

```suggestion
    e.networkUserId = spAnonymous.fold(networkUserId)(_ => "")
```

chuwy (approved):
Looks good. Just couple of questions:

1. Do we consider it bugfix change (a patch-release)? 
2. Do we want to add more unit tests?

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
paulboocock: Nice. I started with the function but since it's the only usage this makes sense to tidy up.

paulboocock:
>     1. Do we consider it a bugfix (a patch-release)?
Yeah I think we should classify this as a bug. It's an oversight as it allows user identifiers to be captured with `SP-Anonymous` present. This fixes the 2.1.x branch which is the recommended branch for Anonymous tracking atm.

>     2. Do we want to add more unit tests?
I've added two new tests to make it clearer why this behaviour exists. It was covered by existing tests but I think the extra documentation of this functionality through further tests is nice to have.

core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala
```diff
@@ -345,10 +345,26 @@ class CollectorServiceSpec extends Specification {
...
         e.contentType shouldEqual ct.get
       }
+      "have an empty nuid if SP-Anonymous is present" in {
+        val l  = `Location`("l")
+        val ct = Some("image/gif")
+        val r  = HttpRequest().withHeaders(l :: hs)
+        val e =
+          service.buildEvent(None, Some("b"), "p", Some("ua"), Some("ref"), "h", "unknown", r, "nuid", ct, Some("*"))
+        e.networkUserId shouldEqual ""
+      }
```
benjben, line 359: You might want to add a test where you add a cookie with a `network_id` in the `HttpRequest`, it seems that you can do it like [that](https://stackoverflow.com/a/43746865/8568080).

benjben (approved):
:+1:

core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala
paulboocock: Managed to add 2 tests at the service level that sets the nuid cookie. The HttpRequest has already been deconstructed by the time we get to the service so I had to split the HttpRequest and HttpCookie into separate params to test it. Perhaps theres a better way(?) but this was the best option I could find whilst keeping the units seperate.

benjben: :+1:

PR #121, opened 2021-02-26
Add SQS collector module (close #120)

### LINKED ISSUE #120: Add SQS collector module

dilyand:
You can run it locally:

```bash
$ sbt "project sqs" "run --config ./app.conf"
```

with `app.conf`:

```
collector {
  interface = "0.0.0.0"
  port = 12345

  paths {}

  p3p {
    policyRef = "/w3c/p3p.xml"
    CP = "NOI DSP COR NID PSA OUR IND COM NAV STA"
  }

  crossDomain {
    enabled = false
    domains = [ "*" ]
    secure = true
  }

  cookie {
    enabled = true
    expiration = "365 days"
    name = sp
    secure = false
    httpOnly = false
    sameSite = "None"
  }

  doNotTrackCookie {
    enabled = false
    name = dnt
    value = 1
  }

  cookieBounce {
    enabled = false
    name = "n3pc"
    fallbackNetworkUserId = "00000000-0000-4000-A000-000000000000"
    forwardedProtocolHeader = "X-Forwarded-Proto"
  }

  enableDefaultRedirect = false

  redirectMacro {
    enabled = false
    placeholder = "[TOKEN]"
  }

  rootResponse {
    enabled = false
    statusCode = 302
  }

  cors {
    accessControlMaxAge = 5 seconds
  }

  prometheusMetrics {
    enabled = false
  }

  streams {
    good = good
    bad = bad
    useIpAddressAsPartitionKey = false

    sink {
      enabled = sqs
      region = "eu-west-1"
      threadPoolSize = 10

      aws {
        accessKey = env
        secretKey = env
      }

      backoffPolicy {
        minBackoff = 1000
        maxBackoff = 10000
      }
    }

    buffer {
      byteLimit = 256000
      recordLimit = 10
      timeLimit = 5000
    }
  }

}

akka {
  loglevel = OFF
  loggers = []

  http.server {
    remote-address-header = on
    raw-request-uri-header = on
    parsing {
      max-uri-length = 32768
      uri-parsing-mode = relaxed
    }
  }
}
```

### DIFF
--- a/.github/workflows/deploy.sh
+++ b/.github/workflows/deploy.sh
@@ -8,7 +8,7 @@ project_version=$(sbt "project core" version -Dsbt.log.noformat=true | perl -ne 
 if [[ "${tag}" = *"${project_version}" ]]; then
   echo "Building and publishing assets (version $project_version)"
   echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
-  sbt "project kinesis" docker:publish && sbt "project pubsub" docker:publish && sbt "project kafka" docker:publish && sbt "project nsq" docker:publish && sbt "project stdout" docker:publish
+  sbt "project kinesis" docker:publish && sbt "project pubsub" docker:publish && sbt "project kafka" docker:publish && sbt "project nsq" docker:publish && sbt "project stdout" docker:publish && sbt "project sqs" docker:publish
 else
   echo "Tag version '${tag}' doesn't match version in scala project ('${project_version}'). Aborting!"
   exit 1
--- a/examples/config.hocon.sample
+++ b/examples/config.hocon.sample
@@ -209,7 +209,10 @@ collector {
     good = {{good}}
     good = ${?COLLECTOR_STREAMS_GOOD}
 
-    # Events that are too big (w.r.t Kinesis 1MB limit) will be stored in the bad stream/topic
+    # Bad rows (https://docs.snowplowanalytics.com/docs/try-snowplow/recipes/recipe-understanding-bad-data/) will be stored in the bad stream/topic.
+    # The collector can currently produce two flavours of bad row:
+    #  - a size_violation if an event is larger that the Kinesis (1MB) or SQS (256KB) limits;
+    #  - a generic_error if a request's querystring cannot be parsed because of illegal characters
     bad = {{bad}}
     bad = ${?COLLECTOR_STREAMS_BAD}
 
@@ -220,14 +223,14 @@ collector {
 
     # Enable the chosen sink by uncommenting the appropriate configuration
     sink {
-      # Choose between kinesis, google-pub-sub, kafka, nsq, or stdout.
+      # Choose between kinesis, sqs, google-pub-sub, kafka, nsq, or stdout.
       # To use stdout, comment or remove everything in the "collector.streams.sink" section except
       # "enabled" which should be set to "stdout".
-      enabled = kinesis
+      enabled = kinesis # or sqs
       enabled = ${?COLLECTOR_STREAMS_SINK_ENABLED}
 
       # Region where the streams are located
-      region = {{kinesisRegion}}
+      region = {{kinesisRegion}} # or {{sqsRegion}}
       region = ${?COLLECTOR_STREAMS_SINK_REGION}
 
       ## Optional endpoint url configuration to override aws kinesis endpoints,
@@ -235,20 +238,22 @@ collector {
       # customEndpoint = {{kinesisEndpoint}}
       # customEndpoint = ${?COLLECTOR_STREAMS_SINK_CUSTOM_ENDPOINT}
 
-      # Thread pool size for Kinesis API requests
+      # Thread pool size for Kinesis and SQS API requests
       threadPoolSize = 10
       threadPoolSize = ${?COLLECTOR_STREAMS_SINK_THREAD_POOL_SIZE}
 
       # Optional SQS buffer for good and bad events (respectively).
       # When messages can't be sent to Kinesis, they will be sent to SQS.
-      # If not configured, sending to Kinesis will be retried. 
+      # If not configured, sending to Kinesis will be retried.
+      # This should only be set up for the Kinesis sink, where it acts as a failsafe.
+      # For the SQS sink, the good and bad queue should be specified under streams.good and streams.bad, respectively and these settings should be ignored.
       sqsGoodBuffer = {{sqsGoodBuffer}}
       sqsGoodBuffer = ${?COLLECTOR_STREAMS_SINK_SQS_GOOD_BUFFER}
 
       sqsBadBuffer = {{sqsBadBuffer}}
       sqsBadBuffer = ${?COLLECTOR_STREAMS_SINK_SQS_BAD_BUFFER}
 
-      # The following are used to authenticate for the Amazon Kinesis sink.
+      # The following are used to authenticate for the Amazon Kinesis and SQS sinks.
       # If both are set to 'default', the default provider chain is used
       # (see http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html)
       # If both are set to 'iam', use AWS IAM Roles to provision credentials.
--- a/sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
+++ b/sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
@@ -1,0 +1,312 @@
+/*
+ * Copyright (c) 2013-2021 Snowplow Analytics Ltd. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package com.snowplowanalytics.snowplow.collectors.scalastream.sinks
+
+import java.nio.ByteBuffer
+import java.util.UUID
+import java.util.concurrent.ScheduledExecutorService
+
+import com.amazonaws.auth.{
+  AWSCredentialsProvider,
+  AWSStaticCredentialsProvider,
+  BasicAWSCredentials,
+  DefaultAWSCredentialsProviderChain,
+  EnvironmentVariableCredentialsProvider,
+  InstanceProfileCredentialsProvider
+}
+import com.amazonaws.services.sqs.{AmazonSQS, AmazonSQSClientBuilder}
+import com.amazonaws.services.sqs.model.{
+  QueueDoesNotExistException,
+  SendMessageBatchRequest,
+  SendMessageBatchRequestEntry
+}
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.{ExecutionContextExecutorService, Future}
+import scala.collection.JavaConverters._
+import scala.concurrent.duration.MILLISECONDS
+import scala.util.{Failure, Random, Success}
+
+import cats.syntax.either._
+
+import com.snowplowanalytics.snowplow.collectors.scalastream.model._
+
+/**
+  *  SQS sink for the Scala Stream Collector
+  */
+class SqsSink private (
+  client: AmazonSQS,
+  sqsConfig: Sqs,
+  bufferConfig: BufferConfig,
+  queueName: String,
+  executorService: ScheduledExecutorService
+) extends Sink {
+  // Records must not exceed 256K when writing to SQS.
+  // Since messages can be base64-encoded, they must meet that limit after encoding.
+  override val MaxBytes = 192000 // 256000 / 4 * 3
+
+  private val ByteThreshold: Long   = bufferConfig.byteLimit
+  private val RecordThreshold: Long = bufferConfig.recordLimit
+  private val TimeThreshold: Long   = bufferConfig.timeLimit
+
+  private val maxBackoff: Long        = sqsConfig.backoffPolicy.maxBackoff
+  private val minBackoff: Long        = sqsConfig.backoffPolicy.minBackoff
+  private val randomGenerator: Random = new java.util.Random()
+
+  implicit lazy val ec: ExecutionContextExecutorService =
+    concurrent.ExecutionContext.fromExecutorService(executorService)
+
+  def storeRawEvents(events: List[Array[Byte]], key: String): List[Array[Byte]] = {
+    events.foreach(e => EventBuffer.store(e))
+    Nil
+  }
+
+  object EventBuffer {
+    type Event = Array[Byte]
+
+    private var storedEvents              = List.empty[ByteBuffer]
+    private var byteCount                 = 0L
+    @volatile private var lastFlushedTime = 0L
+
+    def store(event: Event): Unit = {
+      val eventBytes = ByteBuffer.wrap(event)
+      val eventSize  = eventBytes.capacity
+      if (eventSize >= MaxBytes) {
+        log.error(
+          s"Message of size $eventSize bytes is too large. It must be less than $MaxBytes bytes."
+        )
+      } else {
+        synchronized {
+          if (storedEvents.size + 1 > RecordThreshold || byteCount + eventSize > ByteThreshold) {
+            flush()
+          }
+          storedEvents = eventBytes :: storedEvents
+          byteCount += eventSize
+        }
+      }
+    }
+
+    def flush(): Unit = {
+      val eventsToSend = synchronized {
+        val evts = storedEvents.reverse
+        storedEvents = Nil
+        byteCount    = 0
+        evts.map(_.array())
+      }
+      lastFlushedTime = System.currentTimeMillis()
+      sendBatch(eventsToSend)
+    }
+
+    def getLastFlushTime: Long = lastFlushedTime
+
+    /**
+      * Recursively schedule a task to flush the EventBuffer.
+      * Even if the incoming event flow dries up, all buffered events will eventually get sent.
+      * Whenever TimeThreshold milliseconds have passed since the last call to flush, call flush.
+      * @param interval When to schedule the next flush
+      */
+    def scheduleFlush(interval: Long = TimeThreshold): Unit = {
+      executorService.schedule(
+        new Thread {
+          override def run(): Unit = {
+            val lastFlushed = getLastFlushTime
+            val currentTime = System.currentTimeMillis()
+            if (currentTime - lastFlushed >= TimeThreshold) {
+              flush()
+              scheduleFlush(TimeThreshold)
+            } else {
+              scheduleFlush(TimeThreshold + lastFlushed - currentTime)
+            }
+          }
+        },
+        interval,
+        MILLISECONDS
+      )
+      ()
+    }
+  }
+
+  def sendBatch(batch: List[EventBuffer.Event], nextBackoff: Long = minBackoff): Unit =
+    if (batch.nonEmpty) {
+      log.info(s"Writing ${batch.size} Thrift records to SQS queue ${queueName}")
+
+      multiPut(batch).onComplete {
+        case Success(s) =>
+          log.info(s"Successfully wrote ${batch.size - s.size} out of ${batch.size} messages.")
+          if (s.nonEmpty) {
+            s.foreach { f =>
+              log.error(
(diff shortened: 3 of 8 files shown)


sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+            case (e, m) if (failures.keys.toSet.contains(m.getId)) =>
+              (e, failures.getOrElse(m.getId, BatchResultErrorInfo("unknown code", "unknown message")))
+          }
+        }
+        .toMap
+    }
+
+  def toSqsMessage(event: EventBuffer.Event): SendMessageBatchRequestEntry =
+    // The UUID is not used anywhere but is required by the constructor
+    new SendMessageBatchRequestEntry(UUID.randomUUID.toString, b64Encode(event))
```
dilyand, line 195: Oh, I actually ended up using this UUID. I must change the comment.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+            .getFailed
+            .asScala
+            .toList
+            .map { bree =>
+              (bree.getId, BatchResultErrorInfo(bree.getCode, bree.getMessage))
+            }
+            .toMap
+          // Events to retry and reasons for failure
+          msgGroup.collect {
+            case (e, m) if (failures.keys.toSet.contains(m.getId)) =>
```
benjben, line 186:
```suggestion
            case (e, m) if (failures.contains(m.getId)) =>
```

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+              (e, failures.getOrElse(m.getId, BatchResultErrorInfo("unknown code", "unknown message")))
```
benjben, line 187:
```suggestion
              (e, failures(m.getId))
```
Safe thanks to the test before

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+            scheduleBatch(s.keys.toList, nextBackoff)
+          }
+        case Failure(f) =>
+          log.error("Writing failed.", f)
+          log.error(s"Retrying in $nextBackoff milliseconds...")
+          scheduleBatch(batch, nextBackoff)
+      }
+    }
+
+  // The batch request can return unsuccessful entries. In this case, we return a map of Events to be retried and the reasons they failed to be inserted.
```
benjben, line 162:
```suggestion
/** @return Empty map in case of success, non-empty map of Events to be retried and the reasons why they failed to be inserted in case of problem. */
```

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
benjben: Don't we want to use the event's UUID ?

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+    client.map { c =>
+      val sqsSink = new SqsSink(c, sqsConfig, bufferConfig, queueName, executorService)
+      sqsSink.EventBuffer.scheduleFlush()
+
+      // When the application is shut down try to send all buffered events.
+      Runtime
+        .getRuntime
+        .addShutdownHook(new Thread {
+          override def run(): Unit = {
+            sqsSink.EventBuffer.flush()
```
benjben, line 259: I think that `flush()` should be inside `shutdown()` function

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+    * @return Whether the queue exists
+    */
+  private def sqsQueueExists(client: AmazonSQS, name: String): Unit = {
+    lazy val log = LoggerFactory.getLogger(getClass())
+    try {
+      client.getQueueUrl(name)
+      ()
+    } catch {
+      case _: QueueDoesNotExistException => log.error(s"SQS queue $name does not exist.")
+      case t: Throwable                  => log.error(t.getMessage)
```
benjben, line 306: Don't we want to crash the app when the queue doesn't exist ?

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+        .withClientConfiguration(
+          new ClientConfiguration().withRetryPolicy(
+            new RetryPolicy(
+              new RetryCondition {
+                override def shouldRetry(
+                  originalRequest: AmazonWebServiceRequest,
+                  exception: AmazonClientException,
+                  retriesAttempted: Int
+                ): Boolean =
+                  retriesAttempted < 30 //max allowed by SQS client
```
benjben, line 282: What happens after 30 failures ? We need to at least log an error

benjben (changes requested):

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
dilyand: Yes, we are. The comment was a left over.

.github/workflows/deploy.sh
```diff
@@ -8,7 +8,7 @@ project_version=$(sbt "project core" version -Dsbt.log.noformat=true | perl -ne
 if [[ "${tag}" = *"${project_version}" ]]; then
   echo "Building and publishing assets (version $project_version)"
   echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
-  sbt "project kinesis" docker:publish && sbt "project pubsub" docker:publish && sbt "project kafka" docker:publish && sbt "project nsq" docker:publish && sbt "project stdout" docker:publish
+  sbt "project kinesis" docker:publish && sbt "project pubsub" docker:publish && sbt "project kafka" docker:publish && sbt "project nsq" docker:publish && sbt "project stdout" docker:publish && sbt "project sqs" docker:publish
```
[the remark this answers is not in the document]
dilyand: Nope, good catch, I would have discovered it when I tried to release the RC.

examples/config.hocon.sample
```diff
@@ -220,35 +220,37 @@ collector {
 
     # Enable the chosen sink by uncommenting the appropriate configuration
     sink {
-      # Choose between kinesis, google-pub-sub, kafka, nsq, or stdout.
+      # Choose between kinesis, sqs, google-pub-sub, kafka, nsq, or stdout.
       # To use stdout, comment or remove everything in the "collector.streams.sink" section except
       # "enabled" which should be set to "stdout".
-      enabled = kinesis
+      enabled = kinesis # or sqs
```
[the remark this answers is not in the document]
dilyand: Kinesis and SQS configuration have several pieces in common: they both have an `aws` section for creds, they both have a `bufferConfig`, etc. The other types of collector do not re-use these parts of the config.

examples/config.hocon.sample
```diff
@@ -220,35 +220,37 @@ collector {
...
       enabled = ${?COLLECTOR_STREAMS_SINK_ENABLED}
 
       # Region where the streams are located
-      region = {{kinesisRegion}}
+      region = {{kinesisRegion}} # or {{sqsRegion}}
```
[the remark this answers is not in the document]
dilyand: This setting is only for Kinesis and SQS. I agree that `sinkRegion` is more concise but won't there be a risk that it may confuse people who are trying to setup the pubsub collector for example?

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+
+  def storeRawEvents(events: List[Array[Byte]], key: String): List[Array[Byte]] = {
+    events.foreach(e => EventBuffer.store(e))
+    Nil
+  }
+
+  object EventBuffer {
+    type Event = ByteBuffer
+
+    private var storedEvents              = List.empty[Event]
```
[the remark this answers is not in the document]
dilyand: I gave it a try but couldn't in a pinch find a way to add conditions for when to return the state and restart.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+    private var byteCount                 = 0L
+    @volatile private var lastFlushedTime = 0L
+
+    def store(event: Array[Byte]): Unit = {
+      val eventBytes = ByteBuffer.wrap(event)
+      val eventSize  = eventBytes.capacity
+      if (eventSize >= MaxBytes) {
+        log.error(
+          s"Message of size $eventSize bytes is too large. It must be less than $MaxBytes bytes."
```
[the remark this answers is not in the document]
dilyand: We do. The CollectorService is in charge of that. By the time this function is called, we've already created the bad rows (if any). So for the good sink, we should never see this message. For the bad sink, it might appear in theory, if the bad row we are trying to write is too big. If that is the case, we log this error, to alert us that something is wrong.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -0,0 +1,335 @@
...
+      batch
+        .map { e =>
+          (e, toSqsMessage(e))
+        }
+        .toMap
+        .grouped(MaxSqsBatchSize)
+        .flatMap { msgGroup =>
+          val batchRequest =
+            new SendMessageBatchRequest().withQueueUrl(queueName).withEntries(msgGroup.values.toList.asJava)
+          val response = client.sendMessageBatch(batchRequest)
```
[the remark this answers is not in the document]
dilyand: No, we don't wait.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
dilyand: No, we removed that behaviour previously because in case of outage, you can keep providing more resources and keep events in memory, in the hope that it will get fixed before the collector dies.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
dilyand: I realised we're not actually using these retries for events. We have our own custom retry logic in `sendBatch` and that has unlimited retries. The retry policy defined in the client is only used when we check if the queue exists, which leads to a long delay at the start of the application, while we're verifying that. I'm removing this retry policy altogether.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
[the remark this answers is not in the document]
benjben: > For the bad sink, it might appear in theory, if the bad row we are trying to write is too big

In this case we can resize the size of the payload inside the bad row, so that it's small enough and it can be emitted

> we log this error, to alert us

In this case an input event doesn't have an associated output (either good or bad)

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
benjben: I still see `UUID.randomUUID.toString`, so we're not using the event's UUID

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
benjben: In this scenario the collector is able to receive and buffer events ? Will the initialization be retried until it finally works and send the buffered events ?

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
benjben: > The retry policy defined in the client is only used when we check if the queue exists, which leads to a long delay at the start of the application, while we're verifying that. I'm removing this retry policy altogether.

So what happens if we start the app during an outage ? It can't check the queue existence, doesn't retry and crash ?

dilyand: No, it will try to check the queue exists, fail to assert that, (log an error)[https://github.com/snowplow/stream-collector/blob/2.2.0-rc2/sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala#L282] and continue. If events come in from a tracker, they will start filling memory and that might ultimately kill the collector. A desperate short-term strategy is to keep launching new collector instances, so that the events have somewhere to go. But, in the medium term, the idea is to also have a buffer which, when filled, will tell the tracker to stop sending events until it receives a green light.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
dilyand: Ah, sorry, I misunderstood your question. We don't want to use the event's `event_id` because we might have collisions there (duplicates) and if they are in the same batch, it would confuse what was loaded and what needs to be retried.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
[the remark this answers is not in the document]
dilyand: > In this case we can resize the size of the payload

Yes, actually, that is exactly what we do: https://github.com/snowplow/stream-collector/blob/master/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/utils/SplitBatch.scala#L143

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
benjben: I must be missing something.

> the idea is to also have a buffer which, when filled, will tell the tracker to stop sending events until it receives a green light

We could also just have the collector down and tracker would know straight away that it should stop sending messages because it won't receive 200.

If collector starts during an outage, queue existence check fails but collector stays up and starts receiving messages and buffers them. Then when outage is over, will events from the buffer be sent to SQS ?

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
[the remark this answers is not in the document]
benjben: Also for `generic_error` ? I'm asking but I don't know if bad rows of this type can be very big

benjben (commented):
Hey @dilyand ,

I'll approve the PR once you have marked all my comments as `Resolved`.

Also, don't we want to remove `.grouped(10)` and use `.fold()` instead to never send event/batch of events to SQS that is bigger than 256k (and emit a bad row when one event is bigger) ?

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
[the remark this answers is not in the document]
dilyand: No, we don't have that for `generic_error` but the only way it can get too big is if the query string is massive and I think browser and server limits will kick in way before it becomes an issue.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
dilyand: I decided not to do this now (better to change it in all places) but I'm interested what benefits you see in putting it there.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
dilyand: The initialization does not get retried. The client is initialized (as in, the flush task is scheduled to be executed by it) regardless of whether the destination exists or not. I we try to send an event to an non-existent queue, then it's the job of `sendBatch` to keep retrying. If the queue is then created, then the retry will succeed.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
dilyand: Let's ignore the buffer feature for now. It will be added in a later release.

Right now, you can start the collector during an outage and it will receive events from the tracker. Those events will have nowhere to go and will start filling the collector's memory. Ultimately the collector will crash if the outage is not resolved. However, you can keep spinning up new collector instances (ie, adding more collector memory) in an attempt to buy some time for the outage to be resolved. It's not a sustainable strategy but it is better than not being able to spin up those instances because you crash on no queue.

dilyand:
Hey @benjben, I think I've resolved all of them now. If there's anything outstanding, it's just an oversight.

> don't we want to remove `.grouped(10)` and use `.fold()` instead to never send event/batch of events to SQS that is bigger than 256k

We are never going to send batches that are bigger if we are only using the sqs collector. This is only an issue when we combine the kinesis collector with the sqs buffer, because Kinesis and SQS have different limits but in the config hocon we can only specify one limit.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
benjben: I think that there is no scenario where we would want to shutdown without flushing.

sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
benjben: I missed the fact that SQS client can be initialized successfully even in case of outage. Great !

benjben (approved):
Looks great @dilyand !

> We are never going to send batches that are bigger if we are only using the sqs collector

I thought that we were only checking the size of individual messages but not the size of a batch and that it could be too big ?

dilyand:
> I thought that we were only checking the size of individual messages but not the size of a batch and that it could be too big ?

The collector service does it's own check, based on hardcoded values in each sink. For example, for the kinesis collector it's here: https://github.com/snowplow/stream-collector/blob/2.2.0-rc2/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala#L264-L265 and for the sqs collector, it's here: https://github.com/snowplow/stream-collector/blob/2.2.0-rc2/sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala#L56 . These values are used to split a collector payload that is too big into smaller payloads, each one of which respects the individual record limit.

Then, each sink implements its own batching logic, where we no longer use those hardcoded values, but instead use the values set in this section of the hocon: https://github.com/snowplow/stream-collector/blob/2.2.0-rc2/examples/config.hocon.sample#L316. The `byteLimit` and the `recordLimit` here describe the max size of a batch request to the respective sink.

If we only have the sqs collector, then the hocon can be set up properly to respect the batch limits, while the hardcoded value ensures we respect the individual record limit. However, if both kinesis and sqs are being used, this where things can get messy.

benjben:
I see, thanks for the explanation @dilyand !

PR #124, opened 2021-03-15
Release 2.2.1

dilyand:
The logging in `sinkBatch` can potentially be extracted to make the function a bit less busy. Looking forward to opinions on that.


benjben:
Hey @dilyand , could you link this PR to an issue, so that we can get some context about what we try to achieve please ?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -245,272 +512,33 @@ object KinesisSink {
...
-          if (storedEvents.size >= RecordThreshold || byteCount >= ByteThreshold) {
-            flush()
-          }
-        }
+  def split(
+    batch: List[Events],
+    getByteSize: List[Events] => Int,
+    maxRecords: Int,
+    maxBytes: Int
+  ): List[List[Events]] = {
```
istreeter, line 529: First of all, you can rewrite this with `foldLeft` if you prefer.  It is roughly equivalent to what you wrote, so you can pick whichever you prefer:

```scala
 def split(
    batch: List[Events],
    getByteSize: List[Events] => Int,
    maxRecords: Int,
    maxBytes: Int
  ): List[List[Events]] =
batch.foldLeft(List.empty[List[Events]]) {
  case (Nil, next) => List(List(next))
  case (h :: t, next) if h.size + 1 > maxRecords || getByteSize(next :: h) > maxBytes =>
    List(next) :: h :: t
  case (h :: t, next) => (next :: h) :: t
}
```

Slightly more importantly.... you call `getBatchByteSize` for each event you add.  This is probably inefficient because `getBatchByteSize` allocates many byte buffers each time you call it.  It would be really nice if you can adapt the `foldLeft` so you accumulate the batch size at the same time as folding over the events.

Finally.... I think your `split` returns the events in reverse order.  Does this matter?  Do you un-reverse them before sending to sqs or is it unimportant?

istreeter (commented):
For the most part I think this is a good change.  It is clear and easy to understand even though it is implementing quite a complicated retry algorithm.  I just made a few comments.

dilyand:
@benjben I opened #125.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,323 @@ import cats.syntax.either._
...
+          storedEvents += Events(eventBytes.array(), key)
+          byteCount += eventSize
+        }
+      }
+    }
+
+    def flush(): Unit = {
+      val eventsToSend = synchronized {
+        val evts = storedEvents
+        storedEvents.clear()
```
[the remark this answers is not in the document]
dilyand: Yes, good catch!

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+    * @param target A Kinesis stream or SQS queue to write events to
+    * @param nextBackoff Period of time to wait before next attempt if retrying
+    * @param retriesLeft How many retries are left to the current target. Once exhausted, the target will be switched.
+    */
+  def sinkBatch(batch: List[Events], target: Target, nextBackoff: Long = minBackoff)(
+    retriesLeft: Int                                                   = 10
+  ): Unit =
+    if (batch.nonEmpty) target match {
+      case KinesisStream(buffer) =>
+        log.info(s"Writing ${batch.size} Thrift records to Kinesis stream $streamName.")
```
benjben, line 179: I don't think that we need to add the name of the stream to each line of log

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+  final case class BatchResultErrorInfo(code: String, message: String)
+
+  /**
+    * Events to be written to Kinesis or SQS.
+    * @param payloads Serialized events extracted from a CollectorPayload.
+    *                 The size of this collection is limited by MaxBytes.
+    *                 Not to be confused with a 'batch' events to sink.
+    * @param key Partition key for Kinesis
+    */
+  final case class Events(payloads: Array[Byte], key: String)
```
benjben, line 356: Why is it plural ?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+  def toSqsMessages(events: List[Events]): List[(Events, SendMessageBatchRequestEntry)] =
+    events.map(e => (e, new SendMessageBatchRequestEntry(UUID.randomUUID.toString, b64Encode(e.payloads))))
+
+  def b64Encode(msg: Array[Byte]): String = {
+    val buffer = java.util.Base64.getEncoder.encode(msg)
+    new String(buffer)
+  }
+
+  def scheduleWrite(batch: List[Events], target: Target, lastBackoff: Long = minBackoff)(
+    retriesLeft: Int
```
benjben, line 300: Why separating this parameter from the other ones ?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+  /**
+    * The number of retries is unlimited, but targets switch every 10 attempts if an SQS buffer is configured.
+    * @param batch A batch of events to be written
+    * @param target A Kinesis stream or SQS queue to write events to
+    * @param nextBackoff Period of time to wait before next attempt if retrying
+    * @param retriesLeft How many retries are left to the current target. Once exhausted, the target will be switched.
+    */
+  def sinkBatch(batch: List[Events], target: Target, nextBackoff: Long = minBackoff)(
+    retriesLeft: Int                                                   = 10
+  ): Unit =
```
benjben, line 176:
```suggestion
  def sinkBatch(batch: List[Events], target: Target, nextBackoff: Long = minBackoff, retriesLeft: Int): Unit =
```

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+          prre.setData(ByteBuffer.wrap(event.payloads))
+          prre
+        }
+        prr.setRecords(putRecordsRequestEntryList.asJava)
+        prr
+      }
+      client.putRecords(putRecordsRequest)
+    }
+
+  def writeBatchToSqs(batch: List[Events], sqs: SqsClientAndName): Future[List[(Events, BatchResultErrorInfo)]] =
```
benjben, line 267: Would be nice to add a comment to say that this function returns failed inserts.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+                s"Retrying writing batch to Kinesis stream $streamName in $nextBackoff milliseconds..."
+              else
+                s"Sending batch to SQS queue ${buffer.get.sqsBufferName} in $nextBackoff milliseconds..."
+            }
+            log.error(errorMessage)
+            scheduleWrite(batch, target, nextBackoff)(retriesLeft - 1)
+        }
+
+      case SqsQueue(sqs) =>
+        log.info(s"Writing ${batch.size} Thrift records to SQS queue ${sqs.sqsBufferName}.")
```
benjben, line 216: I don't think that we need to add the name of the stream to each line of log

benjben (commented):
Looks much better than what we had before @dilyand !

It would be great to split `KinesisSink` into smaller files, e.g.
- `KinesisSink` with the initialization and retry logic
- `KinesisEmitter` with everything related to writing to Kinesis
- `SqsEmitter` with everything related to writing to SQS
- `EventStorage` with the buffer

Did you manage to test the switching mechanism ?

Not related to this PR but now that we have the `split` function it would be great to add it to SQS collector, so that we don't rely on humans always configuring the limits right.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+    ()
+  }
+
+  object EventStorage {
+    private val storedEvents              = ListBuffer.empty[Events]
+    private var byteCount                 = 0L
+    @volatile private var lastFlushedTime = 0L
+
+    def store(event: Array[Byte], key: String): Unit = {
+      val eventBytes = ByteBuffer.wrap(event)
```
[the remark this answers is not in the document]
dilyand: We convert it to a buffer, just so we can measure its size with `capacity`. Also see the next comment.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+      val eventSize  = eventBytes.capacity
```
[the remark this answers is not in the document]
dilyand: According to the docs:

> A buffer's _capacity_ is the number of elements it contains. 

For multi-byte characters, the length of the string can be different than the size in bytes. For example:

```
import java.nio.ByteBuffer

val event = "外殼, 例, 案件, 案, 殼體, 事件"

event.length // 20
ByteBuffer.wrap(event.getBytes).capacity // 40
```

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+      if (eventSize >= MaxBytes) {
+        log.error(
```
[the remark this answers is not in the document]
dilyand: I'm fairly certain that is right but I didn't have the time to think about how I can get to 100% certain, so I just left it as is.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,322 @@ import cats.syntax.either._
...
+          } else {
+            scheduleFlush(TimeThreshold + lastFlushed - currentTime)
+          }
+        }
+      },
+      interval,
+      MILLISECONDS
+    )
+    ()
+  }
```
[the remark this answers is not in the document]
dilyand: That's a great idea but I'm not sure if I will have time to properly test it's working as expected. It seems the recursive scheduling has done fine so far, so I think we can stick with it a bit longer. I'm adding #128.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
dilyand: It's very helpful in testing though. :)

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
dilyand: Each `Events` object contains potentially multiple events. I found the singular confusing, because it created the impressions we're dealing with events one by one. But actually this represents a batch that was extracted from the CollectorPayload to match the Kinesis / SQS limits.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
dilyand: I feel like it belongs as a last parameter, but we can't vary it's value without foregoing the default value of `lastBackoff` if we put them in the same param list.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,323 @@ import cats.syntax.either._
...
+  private val TimeThreshold   = bufferConfig.timeLimit
+
+  private val maxBackoff      = kinesisConfig.backoffPolicy.maxBackoff
+  private val minBackoff      = kinesisConfig.backoffPolicy.minBackoff
+  private val randomGenerator = new java.util.Random()
+
+  private val MaxSqsBatchSizeN     = 10
+  private val MaxSqsBatchSizeBytes = 256000
+
+  log.info("Creating thread pool of size " + kinesisConfig.threadPoolSize)
```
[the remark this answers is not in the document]
dilyand: I opened #129.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -42,9 +39,323 @@ import cats.syntax.either._
...
+  // When SQS buffer is enabled MaxBytes has to be 256k,
+  // but we encode the message with Base64 for SQS, so the limit drops to 192k
+  private val SqsLimit     = 192000 // 256000 / 4 * 3
+  private val KinesisLimit = 1000000
+  override val MaxBytes    = if (WithBuffer) SqsLimit else KinesisLimit
+  // private val BackoffTime          = 3000L
+
+  private val ByteThreshold   = bufferConfig.byteLimit
+  private val RecordThreshold = bufferConfig.recordLimit
+  private val TimeThreshold   = bufferConfig.timeLimit
```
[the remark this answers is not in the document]
dilyand: I think it's all because of `MaxBytes` which is capitalized in the `Sink` trait. It sort of makes sense, since they are constants, but I see your point.

istreeter (approved):
:)

PR #151, opened 2021-06-18
Add option to restrict Access-Control-Allow-Origin header to certain domains

### LINKED ISSUE #15: Scala Stream Collector: restrict Access-Control-Allow-Origin to a predefined list of domains
This would work as follows:

- provide a list of domains as configuration
- if the origin header from the request is one of the above, send back a `Access-Control-Allow-Origin` with the domain
- if not don't send the header back

pithlessly:
Closes #15.

Adds two options to the config file:

- `collector.cors.restrictDomainsTo`: the list of domains for which the ACAO header will be served. If not provided, it is served for all domains as it did prior to this change. For consistency, this has a corresponding environment variable `COLLECTOR_CORS_RESTRICT_DOMAINS_TO`; however using this is ill-advised as only a single domain can be supplied this way.
- `collector.cors.doRestrictDomains`: a boolean parameter that can be set to `false` to temporarily disable the functionality of `collector.cors.restrictDomainsTo`. Defaults to `true`. Corresponding environment variable is `COLLECTOR_CORS_ACCESS_CONTROL_DO_RESTRICT_DOMAINS`.

I've never worked with Scala or this codebase before, so I'm happy to change anything if it's not idiomatic.

### DIFF
--- a/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
+++ b/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
@@ -147,11 +147,9 @@ class CollectorService(
         val headers = bounceLocationHeader(params, request, config.cookieBounce, bounce) ++
           cookieHeader(request, config.cookieConfig, nuid, doNotTrack, spAnonymous) ++
           cacheControl(pixelExpected) ++
-          List(
-            RawHeader("P3P", "policyref=\"%s\", CP=\"%s\"".format(config.p3p.policyRef, config.p3p.CP)),
-            accessControlAllowOriginHeader(request),
-            `Access-Control-Allow-Credentials`(true)
-          )
+          List(RawHeader("P3P", "policyref=\"%s\", CP=\"%s\"".format(config.p3p.policyRef, config.p3p.CP))) ++
+          accessControlAllowOriginHeader(request, config.cors.restrictDomainsToConfig) ++
+          List(`Access-Control-Allow-Credentials`(true))
 
         val (httpResponse, badRedirectResponses) =
           buildHttpResponse(event, params, headers.toList, redirect, pixelExpected, bounce, config.redirectMacro)
@@ -183,8 +181,8 @@ class CollectorService(
 
   def preflightResponse(request: HttpRequest, corsConfig: CORSConfig): HttpResponse =
     HttpResponse().withHeaders(
+      accessControlAllowOriginHeader(request, corsConfig.restrictDomainsToConfig) ++
       List(
-        accessControlAllowOriginHeader(request),
         `Access-Control-Allow-Credentials`(true),
         `Access-Control-Allow-Headers`("Content-Type", "SP-Anonymous"),
         `Access-Control-Max-Age`(corsConfig.accessControlMaxAge.toSeconds)
@@ -477,14 +475,25 @@ class CollectorService(
     * Creates an Access-Control-Allow-Origin header which specifically allows the domain which made
     * the request
     * @param request Incoming request
-    * @return Header allowing only the domain which made the request or everything
+    * @param domainRestriction A list of domains which the ACAO header will only be served for, or `None`
+    * if there are no restrictions.
+    * @return A singleton list containing the specified header. If it violates the restrictions, the list
+    * will be empty. If the request had no `Origin` header, the response header will permit every domain.
     */
-  def accessControlAllowOriginHeader(request: HttpRequest): HttpHeader =
-    `Access-Control-Allow-Origin`(request.headers.find {
-      case `Origin`(_) => true
-      case _           => false
+  def accessControlAllowOriginHeader(
+    request: HttpRequest,
+    restrictDomainsTo: Option[List[String]],
+  ): List[HttpHeader] =
+    request.headers.collectFirst {
+      case `Origin`(origins) => origins
     } match {
-      case Some(`Origin`(origin)) => HttpOriginRange.Default(origin)
-      case _                      => HttpOriginRange.`*`
-    })
+      case None         => List(`Access-Control-Allow-Origin`(HttpOriginRange.`*`))
+      case Some(origins) => {
+        val hosts: Seq[String] = origins.map(_.host.value)
+        if (restrictDomainsTo.forall(hosts.intersect(_).nonEmpty))
+          List(`Access-Control-Allow-Origin`(HttpOriginRange.Default(origins)))
+        else
+          Nil
+      }
+    }
 }
--- a/core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala
+++ b/core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala
@@ -738,16 +738,35 @@ class CollectorServiceSpec extends Specification {
     }
 
     "accessControlAllowOriginHeader" in {
+      val origin1 = HttpOrigin("http", Host("origin1"))
       "give a restricted ACAO header if there is an Origin header in the request" in {
-        val origin  = HttpOrigin("http", Host("origin"))
-        val request = HttpRequest().withHeaders(`Origin`(origin))
-        service.accessControlAllowOriginHeader(request) shouldEqual
-          `Access-Control-Allow-Origin`(HttpOriginRange.Default(List(origin)))
+        val request = HttpRequest().withHeaders(`Origin`(origin1))
+        val header = `Access-Control-Allow-Origin`(HttpOriginRange.Default(List(origin1)))
+        "and no domain restrictions" in {
+          service.accessControlAllowOriginHeader(request, None) shouldEqual List(header)
+        }
+        "and a domain restriction which permits that origin" in {
+          service.accessControlAllowOriginHeader(request, Some(List("origin2", "origin1"))) shouldEqual List(header)
+        }
       }
       "give an open ACAO header if there are no Origin headers in the request" in {
         val request = HttpRequest()
-        service.accessControlAllowOriginHeader(request) shouldEqual
-          `Access-Control-Allow-Origin`(HttpOriginRange.`*`)
+        val header = `Access-Control-Allow-Origin`(HttpOriginRange.`*`)
+        "and no domain restrictions" in {
+          service.accessControlAllowOriginHeader(request, None) shouldEqual List(header)
+        }
+        "and any domain restrictions" in {
+          service.accessControlAllowOriginHeader(request, Some(Nil)) shouldEqual List(header)
+        }
+      }
+      "give no ACAO header if the request's has an Origin header" in {
+        val request = HttpRequest().withHeaders(`Origin`(origin1))
+        "but it does not meet the restrictions" in {
+          service.accessControlAllowOriginHeader(request, Some(List("origin2"))) shouldEqual Nil
+        }
+        "but the restrictions exclude all domains" in {
+          service.accessControlAllowOriginHeader(request, Some(Nil)) shouldEqual Nil
+        }
       }
     }
 
--- a/examples/config.hocon.sample
+++ b/examples/config.hocon.sample
@@ -193,6 +193,14 @@ collector {
     # request can be cached. -1 seconds disables the cache. Chromium max is 10m, Firefox is 24h.
     accessControlMaxAge = 5 seconds
     accessControlMaxAge = ${?COLLECTOR_CORS_ACCESS_CONTROL_MAX_AGE}
+
+    # Whether or not the value in `restrictDomainsTo` is respected. Optional, defaults to true.
+    doRestrictDomains = false
+    doRestrictDomains = ${?COLLECTOR_CORS_ACCESS_CONTROL_DO_RESTRICT_DOMAINS}
+    # An optional list of domains to return the Access-Control-Allow-Origin header for. The header will
+    # not be returned for other domains. If this is not provided, it will be provided for all domains.
+    restrictDomainsTo = [ "example.com" ]
+    restrictDomainsTo = [ ${?COLLECTOR_CORS_RESTRICT_DOMAINS_TO} ]
   }
 
   # Configuration of prometheus http metrics
--- a/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
+++ b/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
@@ -93,7 +93,13 @@ package model {
   )
   final case class P3PConfig(policyRef: String, CP: String)
   final case class CrossDomainConfig(enabled: Boolean, domains: List[String], secure: Boolean)
-  final case class CORSConfig(accessControlMaxAge: FiniteDuration)
+  final case class CORSConfig(
+    accessControlMaxAge: FiniteDuration,
+    doRestrictDomains: Boolean              = true,
+    restrictDomainsTo: Option[List[String]] = None,
+  ) {
+    val restrictDomainsToConfig = if (doRestrictDomains) restrictDomainsTo else None
+  }
   final case class KinesisBackoffPolicyConfig(minBackoff: Long, maxBackoff: Long)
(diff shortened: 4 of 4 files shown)


dilyand:
Many thanks for this @pithlessly ! 

Would you be happy to sign our CLA? Usually, you should have been asked this question automatically when you submitted the PR but our CLA bot needed a bit of debugging. Please bear with us until that's fixed.

Once we have the CLA signed, we can slate this for review.

pithlessly:
Thanks, I filled out the form that was linked on your website docs. Are there any other steps I should take?

istreeter:
Hi @pithlessly - thanks a lot for contributing this PR

I am really interested to hear why you find this CORS change is needed.  I know that it is typically good practice for web resources to only be available on certain domains, in order to stop other websites using an API.  However, I never considered that a requirement for the collector, whose purpose is only to catch all events, but does not provide any "resource" (as in the R in CORS).  We have never yet found a reason to protect the collector from use by 3rd party sites.

I do not have a strong object to this change, but it would be great if we could understand better the reasoning.  It might help us in future to make informed design decisions about the collector.

Sorry I have not had a chance to look deeply at your code but, as @dilyand already said.... we will make sure somebody reviews it properly soon.

timolinjs:
Hi @istreeter / @pithlessly  - I have been following this issue. Our org did request that we implement some basic level of "authentication" for our clickstream collecter.  I suppose the concern would be posting bogus data or a denial of service attack. I agree these scenarios would be highly (extremely) unlikely.  However, the fact that this is an open endpoint always gets flagged during pen tests and requires convincing people (sometimes unsuccessfully) that this is low risk.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -477,14 +475,25 @@ class CollectorService(
...
+    request: HttpRequest,
+    restrictDomainsTo: Option[List[String]],
+  ): List[HttpHeader] =
+    request.headers.collectFirst {
+      case `Origin`(origins) => origins
     } match {
-      case Some(`Origin`(origin)) => HttpOriginRange.Default(origin)
-      case _                      => HttpOriginRange.`*`
-    })
+      case None         => List(`Access-Control-Allow-Origin`(HttpOriginRange.`*`))
```
dilyand, line 490: If users have provided a list of trusted domains, should we not allow them access even if there is no `Origin` header?

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -477,14 +475,25 @@ class CollectorService(
...
+      case Some(origins) => {
+        val hosts: Seq[String] = origins.map(_.host.value)
+        if (restrictDomainsTo.forall(hosts.intersect(_).nonEmpty))
```
dilyand, line 493: Along similar lines like my comment on `case None` above: what is the benefit of this check? The config should be our source of truth as to what domains should be allowed, regardless of whether they are listed in the `Origin` header or not.

Am I missing a use-case here, where we want to further narrow down the list of allowed domains based on origin?

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -477,14 +475,25 @@ class CollectorService(
...
+    * will be empty. If the request had no `Origin` header, the response header will permit every domain.
     */
-  def accessControlAllowOriginHeader(request: HttpRequest): HttpHeader =
-    `Access-Control-Allow-Origin`(request.headers.find {
-      case `Origin`(_) => true
-      case _           => false
+  def accessControlAllowOriginHeader(
+    request: HttpRequest,
+    restrictDomainsTo: Option[List[String]],
+  ): List[HttpHeader] =
```
dilyand, line 486: Do we need to change the return type to a `List` here? What is the downside of returning `Access-Control-Allow-Origin(HttpOriginRange.*)` in case all domains should be allowed and `Access-Control-Allow-Origin(HttpOriginRange.Default(domains))` if only certain domains should be allowed?

core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala
```diff
@@ -738,16 +738,35 @@ class CollectorServiceSpec extends Specification {
...
       }
       "give an open ACAO header if there are no Origin headers in the request" in {
         val request = HttpRequest()
-        service.accessControlAllowOriginHeader(request) shouldEqual
-          `Access-Control-Allow-Origin`(HttpOriginRange.`*`)
+        val header = `Access-Control-Allow-Origin`(HttpOriginRange.`*`)
+        "and no domain restrictions" in {
+          service.accessControlAllowOriginHeader(request, None) shouldEqual List(header)
+        }
+        "and any domain restrictions" in {
```
dilyand, line 758: Is this what we want? If there are domain restrictions but no `Origin` header, should we not still enforce the restrictions?

core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala
```diff
@@ -738,16 +738,35 @@ class CollectorServiceSpec extends Specification {
...
+          service.accessControlAllowOriginHeader(request, Some(Nil)) shouldEqual List(header)
+        }
+      }
+      "give no ACAO header if the request's has an Origin header" in {
```
dilyand, line 762: This is a change from always returning an ACAO header to sometimes not returning one. I wonder what the implications are and whether there are any good use cases that justify this.

dilyand (changes requested):
Thank you for bearing with me while I was reviewing this PR @pithlessly . I've left some comments and would appreciate your opinion and feedback on them.

Please check the individual comments for details but the three main highlights are:
- I think the functionality we're shooting for can be achieved with a simplified configuration that only adds one additional option to the config.
- There is a change in behaviour from always responding with an ACAO header to preflight requests to now sometimes responding with no ACAO header. Is that by design?
- If the user has provided a list of allowed domains, should we not enforce that list regardless of other sources, such as the request's `Origin` header?

As already noted in the comments, we are happy to add this functionality to the collector, since there seems to be some demand for it. But we would like to understand better how it would be used. Perhaps some of the questions above can find their answers in specific scenarios we haven't thought about?

examples/config.hocon.sample
```diff
@@ -193,6 +193,14 @@ collector {
...
     accessControlMaxAge = 5 seconds
     accessControlMaxAge = ${?COLLECTOR_CORS_ACCESS_CONTROL_MAX_AGE}
+
+    # Whether or not the value in `restrictDomainsTo` is respected. Optional, defaults to true.
+    doRestrictDomains = false
+    doRestrictDomains = ${?COLLECTOR_CORS_ACCESS_CONTROL_DO_RESTRICT_DOMAINS}
+    # An optional list of domains to return the Access-Control-Allow-Origin header for. The header will
+    # not be returned for other domains. If this is not provided, it will be provided for all domains.
+    restrictDomainsTo = [ "example.com" ]
+    restrictDomainsTo = [ ${?COLLECTOR_CORS_RESTRICT_DOMAINS_TO} ]
```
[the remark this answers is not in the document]
pithlessly: This makes sense, I hadn't noticed that that pattern was being used elsewhere in the file.

examples/config.hocon.sample
```diff
@@ -193,6 +193,14 @@ collector {
     # request can be cached. -1 seconds disables the cache. Chromium max is 10m, Firefox is 24h.
     accessControlMaxAge = 5 seconds
     accessControlMaxAge = ${?COLLECTOR_CORS_ACCESS_CONTROL_MAX_AGE}
+
+    # Whether or not the value in `restrictDomainsTo` is respected. Optional, defaults to true.
```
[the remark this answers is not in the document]
pithlessly: If `restrictDomainsTo` is omitted the default behavior is to allow all domains. Right now, with an empty list the behavior would be to disallow all domains, which I don't think would ever be useful. I see what you mean when you say we could reduce the possibility of misconfiguration by saying that an omitted list and an empty list both allow all domains. I didn't do this because I was worried it could also break user expectations (e.g. that removing a domain from the list could never make the system more permissive).

However, we can go through with your idea if you think the trade-off makes sense. If we do this I agree that `restrictedDomins` would be a better name.

I want to explain the motivation for the `doRestrictDomains` parameter. I saw that many of the other sub-configurations have a `enable` or `enabled` boolean parameter that just allows the user to turn those variables on or off. I wanted the new configuration to have a similar option, but naming it `enabled` would have been misleading since it doesn't have any effect on the behavior of the other parameter in `cors` that was already there (`accessControlMaxAge`).

Being able to disable fields with a bool parameter has the advantage of not requiring the user to delete the existing contents (for example, perhaps the user is repeatedly modifying the config, and wants to disable the behavior of the field temporarily for testing purposes, but intends to restore it later). I suppose it could also be accomplished by commenting out the relevant lines, but then the commented parts would not be checked for syntax or type errors. This is also why it defaults to `true` — having the functionality of `restrictDomainsTo` disabled by default would probably be more confusing.

So it's true that `doRestrictDomains` is not strictly necessary, even with the current design, it was just there because I thought it would be more consistent with the `enabled` fields elsewhere in the file. Again, I'll defer to you if you don't think this is valuable.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
pithlessly: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#examples), the HTTP spec doesn't provide a way to respond with multiple domains in the `Access-Control-Allow-Origin` header. Instead, the recommended way is to check against the `Origin` header and respond with only that origin if it is found to be valid. I am not entirely sure why Akka's `HttpOriginRange` class expects a `Seq<>` rather than a single element (probably because that class is expected elsewhere in contexts where that would make sense) but it wouldn't be correct here.

The original code operated under the assumption that the collector should always allow the origin. Thus it returned `*` for the ACAO in cases where there was no origin specified in the request, and if an origin was specified, it would return that for the ACAO (since returning `*` in that case would be illegal).

The goal of the change was for this behavior to be preserved as long as the user does not specify any domain restrictions. If restrictions are specified, things look a little different:

- If there is an `Origin` header, we can check its value against the allowed ones. If it doesn't match, not sending the header is the correct behavior and was [recommended by the original creator of the issue](https://github.com/snowplow/stream-collector/issues/15#issue-631619350). Not sending the header will cause the browser to deny access to the response as a default.

- If there is no `Origin` header, we ignore the domain restrictions and keep returning `*`, just as the original version did. There is nothing we can really do about non-cross-origin requests. Perhaps we could check for the special case where the config specifies exactly one domain and return it in this case, but I'm skeptical of whether that would be worth it.

Hopefully this addresses your comments on the implementation/tests and I haven't missed anything.

examples/config.hocon.sample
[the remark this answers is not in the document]
dilyand: Yes, I believe a single option that lets users pass in a list of domains would be best.

We have three cases:
- We want to allow one of more domains (but in the event only one of them will be returned in the ACAO header because of the limitations of how it works -- `ACAO(origin)`)
- We want to allow all domains (`*`)
- We want to disallow all domains (no ACAO header).

This is tricky to express unambiguously in the config but I think the best would be:

```
# Allow one or more domains
allowedDomains = [ "myfirstdomain.com", "myseconddomain.com", ...]

# Disallow all domains
allowedDomains = [ ]

# Allow all domains by deleting the option
# allowedDomains
```

This is still a bit finicky, with having to delete the option to allow `*` but I think it's preferable to something like  `restrictedDomains`. With that, we'd still need to delete the option to allow `*` but we'll also have `restrictedDomains  = [ ]` meaning that no domains will be allowed, while it looks like no domains will be restricted.

I understand your motivation for `doRestrict` and appreciate the effort to make the config consistent. I think in this case the benefits of simplicity outweigh the downsides.

dilyand (changes requested):
Thank you for the explanations @pithlessly . They helped me understand the changes much better. Please see my comments. This proved way trickier than it seemed at first glance but I think we're getting there.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
istreeter: **Regarding point 1**

> The Origin header can contain multiple domains

Well.... the akka type signature makes it look like that's true.  But I cannot find any evidence or documentation to support that.  My understanding is that browser CORS requests contain a single origin.

> with the changes proposed at the moment, we're still sending back all of the Origin domains

Technically Dilyan is right.... but I see this is a non-issue in reality.  There will only ever be one origin in a browser's CORS request.  (The [example with cookie domains](https://github.com/snowplow/stream-collector/blob/master/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala#L408-L428) is a slightly different case because it is has to match subdomains, not just exact matches).

**Regarding point 2**

> There's an outstanding question what to do with:
> `allowedDomains = [ "first.example.com", "second.example.com" ]`

^^ That is an extremely common use case.  A company often has multiple websites on multiple domains that all want to send events to the same collector.  The desired behaviour should be: Allow origins from _either_ `first.example.com` _or_ `second.example.com`.  Therefore I disagree with the current implementation, which requires all configured domains to match the request domain:

```scala
restrictDomainsTo.forall(hosts.intersect(_).nonEmpty)
```

should be

```scala
restrictDomainsTo.exists(hosts.contains(_))
```

I also disagree with Dilyan's suggestion that we return the first domain from the list of allowed domains.  The correct behaviour is to return the domain from the request header, because that is what the browser expects to receive.

core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala
istreeter: I slightly agree.  Having said that, this is testing a non-CORS request so really it doesn't matter what ACAO header is in the response.

core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala
istreeter: True.... but I am satisfied this is opt-in behaviour that only comes with turning on the new feature.  It is unusual to return a ACAO header to a non-CORS request, so this is bringing the collector more in line with "normal" behaviour.

examples/config.hocon.sample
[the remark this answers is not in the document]
istreeter: I completely agree with @dilyand on this one

istreeter (commented):
This PR currently solves the problem of rejecting `POST` requests originating from a browser, e.g. from the [snowplow javascript tracker](https://github.com/snowplow/snowplow-javascript-tracker/tree/master/trackers/browser-tracker).  For `POST` requests with a json content type, the browser would send a pre-flight `OPTIONS` request, and CORS would block the follow-up `POST` request.

However, the javascript tracker can [also be configured to send `GET` requests](https://github.com/snowplow/snowplow-javascript-tracker/blob/3.1.4/libraries/browser-tracker-core/src/tracker/types.ts#L49).  We should also consider [the pixel tracker](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/pixel-tracker/).  Such requests would not be pre-flighted by the browser.  Currently, the collector would accept these events (write them to the sink), and only later reject them to the browser via cors.

Furthermore, the javascript tracker will keep on sending the same event over and over, even though the collector sinks the event to the output topic each time.

If we are really going down this route of restricting events to certain origins, I think it only makes sense to block the request earlier.  i.e. we need a code change so the request is not written to the output topic.  Without that change, it just does not make sense to have any kind of restricted CORS.

Issue #152, opened 2021-08-10
Wrong endpoint for chinese region cn-northwest-1


bageren:
https://github.com/snowplow/stream-collector/blob/b253b68c28cd89d0f155e925d79318b84a821c17/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala#L116-L119

There currently is one more chinese region: cn-northwest-1

If the region is set to that it will call the wrong endpoint, unless you pass a custom one.

PR #156, opened 2021-08-10
Critical bumps

istreeter:
This is a minimal PR patch release to address security vulnerabilities only.  Although I also included #152 because it is small and easy.


benjben (approved):
:+1:

PR #160, opened 2021-08-25
Fix-up some AWS Auth Chain Issues

mmathias01:
Hey all!

This is a PR for the collector that fixes the same issue mentioned in: [this PR for Enrich](https://github.com/snowplow/enrich/pull/490)

I also fixed some of the Docker settings so that they matched they matched the docs, and so that the container wasn't running with an unknown user (caused some interesting permissions issues on attached volumes for me)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -56,7 +56,9 @@ lazy val buildSettings = Seq(
 lazy val dockerSettings = Seq(
   Docker / maintainer := "Snowplow Analytics Ltd. <support@snowplowanalytics.com>",
   dockerBaseImage := "snowplow/base-debian:0.2.2",
+  Docker / daemonUserUid := None, 
   Docker / daemonUser := "snowplow",
+  Docker / defaultLinuxInstallLocation := "/snowplow",
   dockerUpdateLatest := true
 )
 
@@ -93,6 +95,7 @@ lazy val kinesis = project
   .settings(
     libraryDependencies ++= Seq(
       Dependencies.Libraries.kinesis,
+      Dependencies.Libraries.sts,
       Dependencies.Libraries.cbor,
       Dependencies.Libraries.sqs
     )
@@ -107,6 +110,7 @@ lazy val sqs = project
   .settings(
     libraryDependencies ++= Seq(
       Dependencies.Libraries.sqs,
+      Dependencies.Libraries.sts,
       Dependencies.Libraries.cbor
     )
   )
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -52,6 +52,7 @@ object Dependencies {
     val thrift           = "org.apache.thrift"                % "libthrift"               % V.thrift
     val kinesis          = "com.amazonaws"                    % "aws-java-sdk-kinesis"    % V.awsSdk
     val sqs              = "com.amazonaws"                    % "aws-java-sdk-sqs"        % V.awsSdk
+    val sts              = "com.amazonaws"                    % "aws-java-sdk-sts"        % V.awsSdk
     val pubsub           = "com.google.cloud"                 % "google-cloud-pubsub"     % V.pubsub
     val kafkaClients     = "org.apache.kafka"                 % "kafka-clients"           % V.kafka
     val nsqClient        = "com.snowplowanalytics"            % "nsq-java-client"         % V.nsqClient


build.sbt
```diff
@@ -56,7 +56,9 @@ lazy val buildSettings = Seq(
 lazy val dockerSettings = Seq(
   Docker / maintainer := "Snowplow Analytics Ltd. <support@snowplowanalytics.com>",
   dockerBaseImage := "snowplow/base-debian:0.2.2",
+  Docker / daemonUserUid := None,
```
mmathias01, line 59: fix the UID so that it is not unknown

build.sbt
```diff
@@ -56,7 +56,9 @@ lazy val buildSettings = Seq(
 lazy val dockerSettings = Seq(
   Docker / maintainer := "Snowplow Analytics Ltd. <support@snowplowanalytics.com>",
   dockerBaseImage := "snowplow/base-debian:0.2.2",
+  Docker / daemonUserUid := None, 
   Docker / daemonUser := "snowplow",
+  Docker / defaultLinuxInstallLocation := "/snowplow",
```
mmathias01, line 61: output binaries based on docs and docker base image.

build.sbt
```diff
@@ -107,6 +110,7 @@ lazy val sqs = project
   .settings(
     libraryDependencies ++= Seq(
       Dependencies.Libraries.sqs,
+      Dependencies.Libraries.sts,
```
mmathias01, line 113: add STS to fix up AWS default auth chain for K8s

project/Dependencies.scala
```diff
@@ -52,6 +52,7 @@ object Dependencies {
     val thrift           = "org.apache.thrift"                % "libthrift"               % V.thrift
     val kinesis          = "com.amazonaws"                    % "aws-java-sdk-kinesis"    % V.awsSdk
     val sqs              = "com.amazonaws"                    % "aws-java-sdk-sqs"        % V.awsSdk
+    val sts              = "com.amazonaws"                    % "aws-java-sdk-sts"        % V.awsSdk
```
mmathias01, line 55: add STS to Dependency table

mmathias01:
FYI the S3 Loader has the same issue. I can submit a PR for that as well, but I figured if you implement the fixes for the collector and enrich you can probably do the S3 one without me submitting a PR, still I am happy to if you want!

build.sbt
istreeter: I agree this is needed!  But rather than fix it in this commit, I will include it when I merge [this issue](#164) which changes our base image to `adoptopenjdk:11-jre-hotspot-focal`, and fixes various other docker settings.

We are slowly migrating all snowplow applications to use the adoptopenjdk base image, to relieve us of the burden of maintaining our own base image.

build.sbt
istreeter: This is an interesting suggestion.... because as we have been migrating snowplow apps to use a different base image, we have been setting the install location to `/opt/snowplow`.  For example [here in the elasticsearch loader](https://github.com/snowplow/snowplow-elasticsearch-loader/blob/1.0.1/project/BuildSettings.scala#L54) and [here in RDB loader](https://github.com/snowplow/snowplow-rdb-loader/blob/1.2.0/project/BuildSettings.scala#L167).

Do you see any problem if we continue with this trend to put it in `/opt/docker`?  Does this conflict with any documentation you have read?  I might need to go and amend the docs, or provide an upgrade guide.

build.sbt
istreeter: Please can you help me understand this one better?  I am familiar with STS and Kubernetes (although not together!).  But I cannot find any documentation that describes the default auth chain specifically on Kubernetes.

I used [this page](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html) to understand the default auth chain in other contexts, but it does not mention STS.  Do you see an error message if this lib is not included?  What is the error message?  How are you providing credentials to your collector?

I'm not against including this in the next release, if I can understand it better.

istreeter (commented):
Hi @mmathias01 - thanks a lot for your helpful contributions :)

I have added a few questions, to help me understand these changes.  We are planning a small release within the next few days, so there is an opportunity to resolve these questions and get it all merged for you.

build.sbt
mmathias01: totally seems reasonable, I don't think it is actively breaking anything but it does make certain things challenging when connected via console haha.

build.sbt
mmathias01: Actually I think /opt/snowplow is probably a good convention overall. The docs I am referencing are [here](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/stream-collector/setup/) although in looking at those docs again it is really just referencing the config file more than the binaries. Still always good to have the binaries in a consistent / known location. Also it would probably actually make sense to have the config files in a /opt/snowplow/config folder since you cannot mount a single file using K8S easily and mounting a directory removes the previous contents of said directory.

build.sbt
mmathias01: I started to type out a detailed explanation but then I decided it was probably better if I let Amazon explain it which they do [here](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) and [here](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).

Essentially that third option "Web Tokens" allows us to create [Kubernetes Service Accounts](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) that are tied to IAM roles. When those roles are assigned to pods then anything running the AWS API in those pods will use assume that IAM role. This is easily done as part of a standard deployment. The details are spelled out in links above but long story short it uses OIDC, Web Tokens, a service on your K8s cluster, and STS to make it work. 

In the collector container without the STS library (and you have enough levels of debugging enabled) you can actually see a warning in the console that tells you that the "Web Tokens" method was skipped because the STS library is not on the class path. The problem then becomes that container then falls back to the IAM role that the node (EC2 Instance role) has. This is less than ideal for a variety of reasons, not least the lease of which is you don't always want every container running on a node to have access to the same things. This [StackExchange post](https://stackoverflow.com/questions/67175006/snowplow-collector-on-k8s-does-not-use-service-account) spells out the issue pretty well. I was having the same issue which is how I found that post and what set me down the path that ultimately led to this PR.


Hopefully that answers the question, but let me know if you need more :) 

P.S.
If you or anyone are looking for helm charts / manifests for deploying a snowplow pipeline in Kubernetes hit me up, I have a set that works on AWS to deploy the collector, enrich, s3 loader (raw, good, bad), iglu server, postgres loader (enriched, bad), and dataflow runner and am happy to share if they are of interest.

istreeter: That is a very helpful explanation, and helpful links - thank you very much indeed.  I agree we should get this change merged, and we should do something similar in other snowplow apps going forwards.

I think here you have added the lib only to the "sqs" collector, but actually we should add it to the "kinesis" collector also.

build.sbt
istreeter: I think the docs are consistent with the new convention:

- The application code is in `/opt/snowplow`
- The docs suggest to mount configuration files into `/snowplow`

The `/snowplow` directory does not exist in the base image, so if you mount an entire directory to that location then there is no danger of removing previous contents.

istreeter:
This got closed via #170 and associated issue #169 .  Thanks again @mmathias01 for the helpful contribution.

build.sbt
caleb15: > If you or anyone are looking for helm charts / manifests for deploying a snowplow pipeline in Kubernetes hit me up


Yes Please! :D
My email is
c
a
l
e
b
@
1
5
f
i
v
e
.
c
o
m

(email obfuscated to avoid spambots)

PR #170, opened 2021-09-10
Include aws-java-sdk-sts to enable web token authentication

### LINKED ISSUE #169: Include aws-java-sdk-sts to enable web token authentication
This is best described by the [discussion in the original PR](#160)

By including `aws-java-sdk-sts` as a runtime dependency, we allow kubernetes deployments to have fine-grained control over the iam policies on pods.

istreeter:


istreeter:
This is just a rebase of [the original PR](#160) from @mmathias01.  I made the following changes compared to the original PR:

- Removed changes to docker settings, which have been addressed already in another commit
- Added a comment to the sts dependency, so we don't forget and remove it again in a future release!!
- Made it a runtime dependency, not compile time.

oguzhanunlu (approved):
LGTM!

Issue #167, opened 2021-09-09
Addition of telemetry to stream-collector


voropaevp:
Addition of telemetry to stream-collector asset so that we can start to understand some basic information about open source usage to help guide us on where we can improve our OSS:

Application: so that we know which apps are most commonly used by open source users (we want to incorporate telemetry into other Snowplow apps in the future)
App version: so we can start to associate value to a release, and understand how many open source users are upgrading / which versions of our apps are in most use
Heartbeat event: tells us how long an app is running for, so we can understand usage over time
User-provided id: an optional input, but when populated, allows us to tie events together across applications so that we can get an idea of which applications are being used together
Acceptance Criteria:

The schemas for the telemetry events & entity are published to Iglu central
Telemetry will be opt-out, and documentation on docs.snowplowanalytics.com describes how to disable this feature
The configuration examples must include an example of disabling the feature

It must send the following event regularly (hourly):
- application_heartbeat

It must send the following fields in the context:
- Application name
- Application version
- Auto-generated ID
- Cloud
- Region

It must be possible for users to provide the following field, but this is optional:
- User-provided ID

The following fields in the context are optional and will only be provided when an app is deployed using the quick start module (will be passed as environment variable):
- Module version
- Module name

Note: we will always be transparent about what we are tracking and why. You can find more about our telemetry principles on this page.

PR #171, opened 2021-09-16
Add telemetry to stream-collector (#167)

voropaevp:


benjben (commented):
Looks good overall !

The commit should be `Add telemetry (close #167)` so that the issue will get closed automatically when merging into `master`.

.github/workflows/telemetry/sender_config/config.hocon
```diff
@@ -0,0 +1,241 @@
...
+    # - the time in milliseconds since the buffer was last emptied reaches time-limit
+    buffer {
+      byteLimit = 100000
+      recordLimit = 40
+      timeLimit = 1000
+    }
+  }
+
+   telemetry {
+    isDisabled  = false
```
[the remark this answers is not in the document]
voropaevp: It won't. I've put this into the example.hocon. What are the other places it should be mentioned?

.github/workflows/telemetry/sender_config/config.hocon
```diff
@@ -0,0 +1,241 @@
...
+    intervalSec = 1
+    method = "POST"
+    url = "127.0.0.1"
+    port = 9191
+    isSecure = false
+    userProvidedId = userProvidedIdValue
+    moduleName = moduleNameValue
```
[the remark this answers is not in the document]
voropaevp: This is an integration test, not a reference config. Idea was to omit some values to see that case is handled. I'll leave a comment.

core/src/main/scala/com.snowplowanalytics.snowplow.telemetry/TelemetryPayload.scala
```diff
@@ -0,0 +1,14 @@
...
+  userProvidedId: Option[String]     = None,
+  moduleName: Option[String]         = None,
+  moduleVersion: Option[String]      = None,
+  instanceId: Option[String]         = None,
+  region: Option[String]             = None,
+  cloud: Option[String]              = None,
+  applicationName: Option[String]    = None,
+  applicationVersion: Option[String] = None,
+  appGeneratedId: String
+)
```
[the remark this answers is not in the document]
voropaevp: Probably the most important one. Missed it, thanks for noticing!

.github/workflows/telemetry/verify_micro_content.py
```diff
@@ -0,0 +1,56 @@
...
+    #     "data": {
+    #         "schema": "iglu:com.snowplowanalytics.snowplow.telemetry/heartbeat/jsonschema/1-0-0",
+    #         "data": {
+    #             "userProvidedId": "userProvidedIdValue",
+    #             "moduleName": "moduleNameValue",
+    #             "moduleVersion": null,
+    #             "instanceId": null,
+    #             "region": null,
+    #             "cloud": null,
+    #             "applicationName": "snowplow-stream-collector",
```
istreeter, line 26: Ideally `applicationName` should be consistent with the terraform module, [which sets it to "stream-collector"](https://github.com/snowplow-devops/terraform-aws-collector-kinesis-ec2/blob/v0.2.0/main.tf#L5).

I actually think "stream-collector" is not a good name.  Because it is missing the `-kinesis` / `-sqs` / `-kafka` suffix.  (And less importantly, I never liked the "stream-" prefix, but that is more personal opinion).

I suggest you talk to Emily about what is a helpful application name for her.  I guess she will tell you it is helpful for her if terraform and collector send the same application name.

core/src/main/scala/com.snowplowanalytics.snowplow.telemetry/TelemetryConfig.scala
```diff
@@ -0,0 +1,17 @@
+package com.snowplowanalytics.snowplow.telemetry
+
+final case class TelemetryConfig(
+  // General params
+  isDisabled: Boolean = false,
+  intervalSec: Long   = 3600L,
```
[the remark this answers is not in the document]
istreeter: I agree with Ben's comment.  But.... I think the collector needs a big tidy-up of where configuration is set, which should include simplifying the hocon file and start using the `src/main/resources/` for all defaults.  Would we be overloading the current commit if we try to address that here at the same time?

build.sbt
```diff
@@ -43,10 +46,15 @@ lazy val commonExclusions = Seq(
   "org.apache.tomcat.embed" % "tomcat-embed-core" // exclude for security vulnerabilities introduced by libthrift
 )
 
+lazy val buildInfoSettings = Seq(
+  buildInfoPackage := "com.snowplowanalytics.snowplow.collectors.scalastream.subproject.generated",
```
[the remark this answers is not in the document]
voropaevp: This is nessesary due to compile errors. See the following:
https://github.com/voropaevp/stream-collector/runs/3619339598?check_suite_focus=true

build.sbt
```diff
@@ -76,14 +84,8 @@ lazy val core = project
...
-    buildInfoKeys := Seq[BuildInfoKey](
-      organization,
-      name,
-      version,
-      "shortName" -> "ssc",
-      scalaVersion
-    ),
-    buildInfoPackage := "com.snowplowanalytics.snowplow.collectors.scalastream.generated"
+    buildInfoPackage := "com.snowplowanalytics.snowplow.collectors.scalastream.generated",
+    buildInfoKeys := Seq[BuildInfoKey](organization, name, version, "shortName" -> "ssc", scalaVersion)
```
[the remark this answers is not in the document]
voropaevp: https://github.com/snowplow/stream-collector/pull/171#discussion_r711092679

build.sbt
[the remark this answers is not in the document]
benjben: I wonder if `BuildInfo` should be used in `buildSettings`. I feel that only each `Main` module should use it (e.g. nsq, kinesis, pubsub etc.). In this case it should work, without `subproject`.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/CloudVendor.scala
```diff
@@ -0,0 +1,13 @@
+package com.snowplowanalytics.snowplow.collectors.scalastream.telemetry
+
+sealed trait CloudVendor {
+  override def toString: String = this match {
+    case CloudVendor.Aws => "AWS"
+    case CloudVendor.Gcp => "GCP"
+  }
```
benjben, line 7: We probably want to use an encoder here and not use `toString`. See [here](https://github.com/snowplow/enrich/blob/feature/file-sink-rotate/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Telemetry.scala#L121)

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryAkkaService.scala
```diff
@@ -0,0 +1,123 @@
...
+            log.warn(s"Scala telemetry tracker has stopped trying to deliver payload after following failure: $failure")
+        }
+
+      val emitter: SyncEmitter = SyncEmitter(
+        EndpointParams(teleCfg.url, port = teleCfg.port, https = teleCfg.isSecure),
+        callback = Some(emitterCallback)
+      )
+      // telemetry - Unique identifier for website / application (aid)
+      // root - The tracker namespace (tna)
+      val tracker = new Tracker(NonEmptyList.of(emitter), "root", "telemetry")
```
benjben, line 77:
```suggestion
      val tracker = new Tracker(NonEmptyList.of(emitter), "tracker-telemetry", appName)
```

See [here](https://github.com/snowplow/enrich/blob/feature/file-sink-rotate/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Telemetry.scala#L79)

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryPayload.scala
```diff
@@ -0,0 +1,15 @@
+package com.snowplowanalytics.snowplow.collectors.scalastream.telemetry
+
+// iglu:com.snowplowanalytics.oss/oss_context/jsonschema/1-0-1
+private case class TelemetryPayload(
+  userProvidedId: Option[String]  = None,
+  moduleName: Option[String]      = None,
+  moduleVersion: Option[String]   = None,
+  instanceId: Option[String]      = None,
+  region: Option[String]          = None,
+  cloud: Option[String]           = None,
```
benjben, line 10:
```suggestion
  cloud: Option[CloudVendor]           = None,
```

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryPayload.scala
```diff
@@ -0,0 +1,15 @@
+package com.snowplowanalytics.snowplow.collectors.scalastream.telemetry
+
+// iglu:com.snowplowanalytics.oss/oss_context/jsonschema/1-0-1
+private case class TelemetryPayload(
```
benjben, line 4: This class is used only to create the `Json` with `.asJson`. We can get rid of it and create it directly in `makeHeartbeatEvent`, like [here](https://github.com/snowplow/enrich/blob/feature/file-sink-rotate/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Telemetry.scala#L101-L112)

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryAkkaService.scala
```diff
@@ -0,0 +1,123 @@
...
+              s"Scala telemetry tracker failed to reach ${params.getUri} with following exception $exception" +
+                s" after ${req.attempt} attempt"
+            )
+          case Result.RetriesExceeded(failure) =>
+            log.warn(s"Scala telemetry tracker has stopped trying to deliver payload after following failure: $failure")
+        }
+
+      val emitter: SyncEmitter = SyncEmitter(
+        EndpointParams(teleCfg.url, port = teleCfg.port, https = teleCfg.isSecure),
+        callback = Some(emitterCallback)
```
benjben, line 73: Would be good to perform some retries, see [here](https://github.com/snowplow/enrich/blob/feature/file-sink-rotate/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Telemetry.scala#L76)

voropaevp: Defaults to 10 retries, same as the example.

https://github.com/snowplow/snowplow-scala-tracker/blob/af50b940cb6c258f294c2bb878b633a0f42c7228/modules/id-emitter/src/main/scala/com/snowplowanalytics/snowplow/scalatracker/emitters/id/SyncEmitter.scala#L89

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryAkkaService.scala
voropaevp: Reference code sets application id to `root` and tracker namespace to `appName`.  Please elaborate.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryAkkaService.scala
benjben: Good point I missed that

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryAkkaService.scala
benjben: [This is](https://github.com/snowplow/snowplow-scala-tracker/blob/master/modules/core/src/main/scala/com.snowplowanalytics.snowplow/scalatracker/Tracker.scala#L72-L73) how these 2 fields are used in the tracker.

And this is [what these fields mean](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/snowplow-tracker-protocol/#common-parameters-platform-and-event-independent).

I think that all telemetry trackers across apps should use the same namespace and then each app should identify itself.

examples/config.hocon.sample
```diff
@@ -322,7 +322,27 @@ collector {
       timeLimit = ${?COLLECTOR_STREAMS_BUFFER_TIME_LIMIT}
     }
   }
-
+  # Telemetry sends heartbeat events to external pipeline.
+  # Unless isDisabled parameter set to true, this feature will be enabled. Deleting whole section will not disable it.
+  # Schema URI: iglu:com.snowplowanalytics.oss/oss_context/jsonschema/1-0-1
+  #
+  telemetry {
+      isDisabled = false
```
benjben, line 330: We generally prefer to use a verb instead of passive, see [here](https://github.com/snowplow/enrich/blob/feature/file-sink-rotate/config/config.kinesis.reference.hocon#L197)

examples/config.hocon.sample
```diff
@@ -322,7 +322,27 @@ collector {
...
+      interval = 60 minutes
+
+      # Connection properties for the receiving pipeline
+      method = POST
+      url = collector-g.snowplowanalytics.com
+      port = 443
+      isSecure = true
```
benjben, line 337:
```suggestion
      secure = true
```
Same

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryAkkaService.scala
voropaevp: There is already provided the user-defined name for the application, which is a part of the Context.
https://github.com/snowplow/iglu-central/blob/f163f9565c0a91e3dc1c6ae3e56d4ced7f30bae2/schemas/com.snowplowanalytics.oss/oss_context/jsonschema/1-0-1#L76

So suggestion is to use the following conversion:
- aid - telemetry (constant)
- tna - name of component family - collector, enrich

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryPayload.scala
voropaevp: I think it is good practice to have a class just for the single purpose of serializing it. Mostly for documentation reasons, and for theoretical flexibility in case we start to do more things with it. 

There is a single call to toJson, while an example has toJson call for every field.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/CloudVendor.scala
voropaevp: The Collector does not use any typeclasses. The static method fits the style of the project.

voropaevp: Core makes use of the ProjInfo to print the help message. 

https://github.com/voropaevp/stream-collector/blob/9cf5921e21bfdb45b2cb72e2736924c4ef2fb3ea/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala#L45

We could pass this into the parseConfig from the `Main` modules. 
` val (collectorConf, akkaConf) = parseConfig(args, buildInfo....)`

Do you have alternative suggestions?

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryPayload.scala
istreeter: I think both methods are valid, and both sides of the argument are valid too..  I've done it both ways in the past.

I'm even OK with being inconsistent between collector and enrich.  Because in this case the inconsistency does not affect the inputs/outputs of the application.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryAkkaService.scala
istreeter: > I think that all telemetry trackers across apps should use the same namespace and then each app should identify itself.

That sounds sensible.

> There is already provided the user-defined name for the application, which is a part of the Context.

Yes true, but with a difference in meaning.  The value in the context describes the _application we are describing_ whereas the `aid` describes the _application sending the event_.  The distinction is important when terraform sends the event.

Currently the [terraform module sets the aid to "terraform-oss-modules"](https://github.com/snowplow-devops/terraform-snowplow-telemetry/blob/v0.2.0/main.tf#L9) and it sets tna to empty string.  The latter could be improved.

For the collector and for enrich it makes sense for `aid` to be exactly equal to the context's `applicationName`.  Because the app sending the event **is** the app being described.

For tna I slightly prefer `telemetry` rather than `tracker-telemetry`.  Not strongly opinionated.  And fairly sure the people using this data won't mind much either.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/CloudVendor.scala
benjben: I'm confused, this comment has nothing to do with the encoder right ?

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/TelemetryAkkaService.scala
benjben: Let's use `"telemetry"`. `tracker` seems redundant

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/telemetry/CloudVendor.scala
benjben: > We could pass this into the parseConfig from the Main modules

Yes I think that each `main()` should pass the `BuildInfo` to the core. That's what we do in enrich

benjben (approved):
:+1: 

Would be great to not use `""` for app and version in the unit tests

oguzhanunlu:
Closing in favor of https://github.com/snowplow/stream-collector/pull/177

PR #175, opened 2021-09-27
Handle LimitExceededException when testing if stream exists (close #174)

### LINKED ISSUE #174: Handle LimitExceededException when testing if stream exists
We use [the kinesis describeStream endpoint](https://github.com/snowplow/stream-collector/blob/2.3.1/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala#L513) to test at startup whether a stream exists. A aws account has a [global limit of 10 requests per second on this endpoint](https://docs.aws.amazon.com/streams/latest/dev/service-sizes-and-limits.html).  For "busy" accounts, this can too easily be exceeded, which means the app crashes at startup with a [`LimitExceededException`](https://docs.aws.amazon.com/ja_jp/AWSJavaSDK/latest/javadoc/com/amazonaws/services/kinesis/model/LimitExceededException.html)

The suggested changes are *all* of:

1. Catch and ignore the `LimitExceededException` when checking the stream exists.  Probably we should ignore _all_ exceptions, because we want collector to start up even when there is a kinesis outage.
2. Add config option to disable checking the stream exists.  Afterall, we only use it for logging a warning.  For some pipelines we know there is a danger of exceeding the limit, and yet we are confident we have configured the stream name correctly.
3. Same for checking the sqs queue exists.  I could not find a api quota for this endpoint, but even so we should handle _any_ error that occurs, instead of crashing the app here.

istreeter:


kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -172,8 +172,17 @@ class KinesisSink private (
...
+              outage = buffer match {
+                case None => false
+                case Some(SqsClientAndName(sqsClient, bufferName)) =>
+                  sqsQueueExists(sqsClient, bufferName) match {
+                    case Success(true)  => false
+                    case Success(false) => true
+                    case Failure(_)     => true
+                  }
+              }
+            }
```
dilyand, line 185: 💪🏻

dilyand (approved):
LGTM!

voropaevp:
`sqsQueueExists` also exists in sqs collector, I guess that `enableStartupChecks` config option should be added there too.

PR #176, opened 2021-09-29
Feature config file cleanup

voropaevp:
Providing healthy defaults for common configuration parameters in `application.conf`

Example minimal and full configuration for each component type.


snowplowcla:
Thanks for your pull request. Is this your first contribution to a Snowplow open source project? Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

:memo: Please visit https://docs.snowplowanalytics.com/docs/contributing/contributor-license-agreement/ to learn more and sign.

Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

examples/config.hocon.kinesis.minimal
```diff
@@ -0,0 +1,25 @@
...
+  interface = "0.0.0.0"
+  port = 80
+
+  streams {
+    good = "{{good}}"
+    bad = "{{bad}}"
+
+    sink {
+      enabled = kinesis
+      region = "{{region}}"
```
istreeter, line 11: I never liked this `{{ }}` syntax.  It doesn't mean anything, and we don't use it any other example file.

The purpose of these files is to be a demonstrative example of typical values.  I think it is OK to put a real example, e.g. `eu-central-1`.  See [here as an example](https://github.com/snowplow-incubator/snowplow-postgres-loader/blob/0.3.1/config/config.kinesis.minimal.hocon#L6).

Same applies for all other keys where you have `{{ }}`.

examples/config.hocon.kinesis.minimal
```diff
@@ -0,0 +1,25 @@
+collector {
+  interface = "0.0.0.0"
+  port = 80
+
+  streams {
+    good = "{{good}}"
+    bad = "{{bad}}"
+
+    sink {
+      enabled = kinesis
```
istreeter, line 10: I would slightly prefer `sink.enabled` to have a default.  I think you can achieve this by having a `application.conf` file in the kinesis module (and pubsub module, etc) with kinesis-specific defaults, and a `reference.conf` file in the `core` module with all other defaults.  But I never tested if this works.

examples/config.hocon.kinesis.minimal
```diff
@@ -0,0 +1,25 @@
...
+      threadPoolSize = 10
+
+      aws {
+        accessKey = iam
+        secretKey = iam
+      }
+
+      backoffPolicy {
+        minBackoff = 3000
+        maxBackoff = 600000
```
istreeter, line 21: Same, I think we can be opinionated about these values and set sensible defaults.

istreeter (commented):
In other repos we call the files `config.kinesis.minimal.hocon` instead of `config.hocon.kinesis.minimal`.  For me it makes sense for file format `hocon` to be at the end of the file name.

examples/config.hocon.kinesis.minimal
```diff
@@ -0,0 +1,25 @@
...
+  port = 80
+
+  streams {
+    good = "{{good}}"
+    bad = "{{bad}}"
+
+    sink {
+      enabled = kinesis
+      region = "{{region}}"
+      threadPoolSize = 10
```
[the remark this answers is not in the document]
dilyand: `10` is a good default.

examples/config.hocon.kinesis.minimal
dilyand: `3000` and `600000` are good defaults.

examples/config.hocon.pubsub.minimal
```diff
@@ -0,0 +1,21 @@
...
+
+  streams {
+    good = "{{good}}"
+    bad = "{{bad}}"
+
+    sink {
+      enabled = google-pub-sub
+      googleProjectId = "{{googleProjectId}}"
+
+      backoffPolicy {
```
dilyand, line 13: These are good defaults:

```
backoffPolicy {
        minBackoff = 1000
        maxBackoff = 1000
        totalBackoff = 10000
        multiplier = 1
      }
```

examples/config.hocon.sqs.minimal
```diff
@@ -0,0 +1,25 @@
...
+  port = 80
+
+  streams {
+    good = "{{good}}"
+    bad = "{{bad}}"
+
+    sink {
+      enabled = sqs
+      region = "{{region}}"
+      threadPoolSize = 10
```
dilyand, line 12: This is a good default.

examples/config.hocon.sqs.minimal
```diff
@@ -0,0 +1,25 @@
...
+
+      aws {
+        accessKey = iam
+        secretKey = iam
+      }
+
+      backoffPolicy {
```
dilyand, line 19: These are good defaults.

dilyand (commented):
Looks good @voropaevp . I left comments on a few values in the minimal configs where I think we can set sensible defaults via `application.conf`.

And one nit: There are some `no new line at end of file` across the hocons.

examples/config.hocon.pubsub.minimal
voropaevp: There is a conflict between good defaults for SQS/Kinesis and GCP.
Would it be a good idea to put `application.conf` in each component instead of the core to resolve this?

dilyand: Yes, I think it would work the same way as @istreeter suggested for the `enabled` default: an `application.conf` in each module that only enables that particular sink.

core/src/main/resources/application.conf
```diff
@@ -0,0 +1,106 @@
...
+    enabled = false
+    name = "n3pc"
+    fallbackNetworkUserId = "00000000-0000-4000-A000-000000000000"
+  }
+
+  cookie {
+    enabled = true
+    expiration = 365 days
+    name = sp
+    domains = [
```
istreeter, line 24: `domains` has type `Option[List[String]]`.  So I think here you force the default to be `Some(Nil)`.  The problem is, I don't know to override this in the user-provided hocon to set it back to `None` again.  Is there a way?

I think a more sensible default is `None` i.e. not present in the hocon.

core/src/main/resources/application.conf
```diff
@@ -0,0 +1,106 @@
...
+
+    ]
+    fallbackDomain = "acme1.com"
```
istreeter, line 27: Same comment: `fallbackDomain` has type `Option[String]` and here you force it to be `Some("acme1.com")` and I don't know how to override it at run time to make it `None`.

core/src/main/resources/application.conf
```diff
@@ -0,0 +1,106 @@
...
+    secure = true
+    httpOnly = false
+    sameSite = "None"
```
istreeter, line 30: Please @paulboocock can you confirm these are sensible default values for cookies?  Only used if the user does not override them.

core/src/main/resources/application.conf
```diff
@@ -0,0 +1,106 @@
...
+  doNotTrackCookie {
+    enabled = false
+    name = ""
+    value = ""
+  }
+
+  ssl {
+    enable = false
+    redirect = false
+    port = 9543
```
istreeter, line 42: Must be 443 or else [this is a breaking change](https://github.com/snowplow/stream-collector/blob/2.3.1/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala#L150)

core/src/main/resources/application.conf
```diff
@@ -0,0 +1,106 @@
...
+
+  prometheusMetrics {
+    enabled = false
+  }
+
+  rootResponse {
+    enabled = false
+    statusCode = 302
+    headers = {}
+    body = "302, redirecting"
```
istreeter, line 58: Strictly speaking should be [empty string for backwards compatibility](https://github.com/snowplow/stream-collector/blob/2.3.1/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala#L92).  On the other hand, I'm fairly sure nobody ever sees this string.

core/src/main/resources/application.conf
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+
+  redirectMacro {
+    enabled = false
+  }
+
+  cors {
+    accessControlMaxAge = 5 seconds
```
istreeter, line 67: @paulboocock could also comment on this one.

istreeter (commented):
This is great @voropaevp!  To explain my comments..... if a param previously had a default, then we should keep it the same in case a user has a hocon file with the param missing.  Upgrades must be seamless for those users.

If a param previously had no default, then it is OK to add a sensible default.  Because existing hocon files remain valid.

Please also note the new [enableStartupChecks flag](https://github.com/snowplow/stream-collector/blob/fbd5097153c7cc8d83d02be90335f398b8ffb835/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala#L159) which should default to `true` when you rebase your branch.

The default values in the `model.scala` file are now redundant.  E.g. [`port: Int = 443"`](https://github.com/snowplow/stream-collector/blob/2.3.1/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala#L150) can now be `port: Int`, because the defaults come from elsewhere now.  On the other hand they are technically harmless, so your choice whether you remove those or not.

core/src/main/resources/application.conf
paulboocock: These settings are the best defaults for modern browsers, and have the widest support for locations where you would track (such as cross domain, in an iframe, etc.). [Some older browsers](https://www.chromium.org/updates/same-site/incompatible-clients) reject `SameSite = None` cookies, so there is an argument for `Lax` being a better default but that wouldn't work cross domain. In reality, there isn't a great default for `sameSite` but I think for a server set cookie, None is probably the safest default.

As for other fields, `httpOnly` should probably be true by default. There's no reason to read this cookie client side, so not much reason to make it false. `secure` must be true is `SameSite = None`, so thats a good default.

core/src/main/resources/application.conf
paulboocock: I'm not sure why this is so short to be honest, besides 5 seconds being the chromium default. Increasing this would lead to fewer OPTIONS requests being sent and given that a preflight from the collector would be unlikely the change, I wonder if a large value like `600` (10 minutes) would be better. I picked `600` because thats the cap for Chromium < 76. 

Another benefit of increasing this value is that we increase the chance of events being captured on page navigation, since we're not waiting for the OPTIONS request before sending the POST.

Downsides would be having to wait 10 minutes for it to expire if it's misconfigured or if the configuration changes, which could lead to CORS errors.

voropaevp:
see the merged reference.conf:
[reference.conf.txt](https://github.com/snowplow/stream-collector/files/7269000/reference.conf.txt)

It works correctly overriding the parameters from bottom to the top. I have experimented with printing akka configuration in runtime to verify that. Changing    max-connections = 2048 in akka section correctly overwrote runtime value.

voropaevp:
I've prepared a branch that is ready for the merge. Merging this is problematic due to it being in the other repo.
https://github.com/snowplow/stream-collector/tree/feature/simple_config

PR #179, opened 2021-10-08
Ignore warnings raised by invalid user agents

### LINKED ISSUE #178: Suppress warnings for invalid user agent
Currently Akka raises warnings where the User-Agent is considered to have invalid input. There are a large number of useragents including Facebook, bots and just regular useragents that contain these symbols (e.g., [, ], ;, etc) so this can result in a large number of warnings raised in the logs (Cloudwatch / Cloud Monitoring).

In addition - due to the way these errors are raised with newlines this yields two rows (one which is just ^). For higher volumes pipelines this can easily result in tens to hundreds of million warnings per day which accumulate associated costs in terms of network egress / log storage.

As this is a warning these useragents can be processed / enriched without any issue in the Snowplow pipeline so removing them does not cause any side effects.

```
[WARN] [10/08/2021 14:53:36.375] [scala-stream-collector-akka.actor.default-dispatcher-4] [akka.actor.ActorSystemImpl(scala-stream-collector)] Illegal header: Illegal 'user-agent' header: Invalid input '[', expected 'EOI', product-or-comment, WSP, comment, CRLF or ws (line 1, column 6): YEET [123]
     ^
```

miike:
Fixes #178 

Not too sure if I should add this just to the sample hocon or the other hocon as well?

### DIFF
--- a/examples/config.hocon.sample
+++ b/examples/config.hocon.sample
@@ -351,6 +351,7 @@ akka {
       max-uri-length = ${?AKKA_HTTP_SERVER_PARSING_MAX_URI_LENGTH}
       uri-parsing-mode = relaxed
       uri-parsing-mode = ${?AKKA_HTTP_SERVER_PARSING_URI_PARSING_MODE}
+      ignore-illegal-header-for = ["user-agent"]
     }
   }
 


istreeter:
Thanks @miike!  This would certainly make our logs cleaner.

I would consider adding it to the `application.conf` files (e.g. [this one](https://github.com/snowplow/stream-collector/blob/2.4.1/kinesis/src/main/resources/application.conf#L27)) so that it is on by default for all collectors.

We can get this into the next release, whenever that may be.

miike:
Closing in favour of #220.

PR #194, opened 2021-12-08
Release 2.4.2

istreeter:


project/Dependencies.scala
```diff
@@ -40,7 +40,7 @@ object Dependencies {
     // Scala
     val collectorPayload = "0.0.0"
     val tracker          = "1.0.0"
-    val akkaHttp         = "10.2.6"
+    val akkaHttp         = "10.2.7"
```
dilyand, line 43: Just to double-check: the max parsing depth will be configurable using the `akka` section of the config HOCON, correct?

dilyand (approved):
LGTM @istreeter . 

Are we considering this a `hotfix` and hence merging straight to `master`? It seems like we haven't followed the usual `feature > develop > release > master` path.

project/Dependencies.scala
istreeter: It will be configurable, in the same way that all akka options are configurable in the `akka` seciont.  But we will leave it on the akka default, which is 5.  I don't see a reason to change from the default.

istreeter:
Yes I consider this a hotfix.

PR #215, opened 2022-01-10
2.5.0 candidates

istreeter:


core/src/main/resources/reference.conf
```diff
@@ -1,5 +1,9 @@
 enableDefaultRedirect = false
+redirectDomains     = []
 enableStartupChecks = true
+terminationDeadline = 10.seconds
+preTerminationPeriod = 10.seconds
+preTerminationUnhealthy = false
```
chuwy, line 6: Couldn't find any documentation on these. Especially curious why we want `preTerminationUnhealthy` to be `false` by default.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
```diff
@@ -170,5 +174,23 @@ trait Collector {
...
+        if (collectorConf.preTerminationUnhealthy) {
+          log.info("setting health endpoint to unhealthy")
+          health.toUnhealthy()
+        }
+        log.info(s"Sleeping for ${collectorConf.preTerminationPeriod}")
+        Thread.sleep(collectorConf.preTerminationPeriod.toMillis)
+        log.info("Initiating http server termination")
+        Await.result(system.terminate(), collectorConf.terminationDeadline)
+        log.info("Server terminated. Initiating sinks shutdown")
+        sinks.good.shutdown()
```
chuwy, line 191: What if some of hooks block or cannot complete at all - for example `sinkBatch` in `SqsSink` cannot succeed. Could there be a sensible timeout? Could we do `good` and `bad` sinks in parallel (although we do so much scary stuff with threads here already)?  Shouldn't it be part of `terminationDeadline`? What if `Await` fails to succeed - to me it seems that `good.shutdown` is a) more important; b) are we sure all sinks can flush and shutdown gracefully in absence of `system`? 

Another unrelated nit is that we likely want to log something of it with `warn` level - it's an important message and often people set logging severity to warning.

chuwy (commented):
Looks great! I just have some questions about shutdown hook - that part looks quite scary to me, but although we certainly improve the shutdown - I think there are still corner cases that we can be more clear about.

core/src/main/resources/reference.conf
istreeter: There is [some documentation in the example hocon files](https://github.com/snowplow/stream-collector/blob/feature/graceful-shutdown/examples/config.kinesis.extended.hocon#L259-L268).

For why it's false by default.... I have been told that it in a typical kubernetes setup we don't really need the health endpoint to return 503 during shutdown.  It's completely OK for the health endpoint to always return 200, even when shutting down.  I am adding the feature because I am told it can be helpful for debugging shutdown problems.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
istreeter: Really good points.

My preference is to **not** have a timeout on the sinks.  Most orchestration platforms (e.g. kubernetes) will kill a container if it has not completed graceful shutdown within a configurable amount of time.  I think the orchestration platform is the proper place to configure this, rather than the application.

I do, however, agree it would be better to shut down the two sinks in parallel.  And I do agree it needs more error handling, so that sinks are shutdown even if actor system is not successfully terminated.  I will make those changes.

> are we sure all sinks can flush and shutdown gracefully in absence of system

Yes I think we are sure.  Because sinks are unaware that they run in the context of an akka system.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
```diff
@@ -170,5 +177,38 @@ trait Collector {
...
+        log.warn("Initiating sinks shutdown")
+        val shutdowns = List(shutdownSink(sinks.good), shutdownSink(sinks.bad))
+        Await.result(Future.sequence(shutdowns), Duration.Inf)
+        ()
+      }))
   }
+
+  private def shutdownSink(sink: Sink)(implicit ec: ExecutionContext): Future[Unit] =
+    Future {
+      sink.shutdown()
```
chuwy, line 209: Very last thing - when we look at the logs it'd be great to see that sink `A` has flushed successfully (although doesn't seem we can get meaningful names for them)

chuwy (approved):
Looks great!

One nit, but I think it's totally fine to skip if it's more than one line to name those sinks.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
istreeter: I did what I could.  It now says "good sink" or "bad sink" before trying to terminate and after successfully terminating.

PR #220, opened 2022-01-28
Ignore warnings raised by illegal user agents

miike:
Copy of #179 with all the hocon / conf files updated inline to prevent the logging of warnings for invalid user agent (which turns out to be a lot of user agents).

### DIFF
--- a/.github/workflows/integration_tests/telemetry/micro_config/config.hocon
+++ b/.github/workflows/integration_tests/telemetry/micro_config/config.hocon
@@ -79,6 +79,7 @@ akka {
     parsing {
       max-uri-length = 32768
       uri-parsing-mode = relaxed
+      ignore-illegal-header-for = ["user-agent"]
     }
   }
  }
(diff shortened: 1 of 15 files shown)


miike:
cc @istreeter

istreeter:
Hey @miike many many thanks indeed for submitting this fix.

Unfortunately I think it came a couple of days too late.  We ended up preparing [this fix](https://github.com/snowplow/stream-collector/commit/e9b9a616c11787ab3cb6182861ac4221a755c0e1) a couple of weeks ago, and it got merged along with the 2.5.0 release yesterday.

I feel bad that you went to this effort and we ended up not using it.  But.... at least the feature is merged.  I hope all our users appreciate the new cleaner logs!  Thank you again for suggesting this config change.

miike:
No worries @istreeter at all. I think having a fix that prevents additional logging is probably more advantageous given log4j etc so this is a better solution!

Issue #221, opened 2022-02-24
Inspect X-Forwarded-Proto for http -> https redirects


istreeter:
The collector currently listens for unencrypted requests on port 80 and encrypted requests on port 443.  It can be [configured so that port 80 returns 302 redirects to the https endpoint](https://github.com/snowplow/stream-collector/blob/37562c8dedc6aa2e4dc814afd6f2c56f7c394654/examples/config.kinesis.extended.hocon#L28-L29).

However, there are two production setups for which this redirect does not work:
1. If the load balancer is set up to route all traffic to the collector's port 443, including incoming http and https requests.  This setup might be used so that all traffic within the vpc is always encrypted.
2. If the collector is behind a reverse proxy that routes all traffic to the collector's port 80.  This setup might be used if a reverse proxy (e.g. nginx) is being used to terminate TLS.

In both cases, the collector does not see any difference between http and https requests, because they all land on the same port.

The solution is for the collector to inspect the `X-Forwarded-Proto` header, and to issue a redirect if the original request received by the load balancer was http.

PR #224, opened 2022-03-04
Inspect X-Forwarded-Proto for http -> https redirects

oguzhanunlu:


core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
```diff
@@ -155,7 +159,11 @@ trait Collector {
         Http()
           .newServerAt(collectorConf.interface, collectorConf.ssl.port)
           .enableHttps(ConnectionContext.httpsServer(SSLContext.getDefault)),
-        routes
+        optionalHeaderValueByName("X-Forwarded-Proto") {
+          case Some(clientProtocol) if clientProtocol == "http" =>
+            redirectTo(Scheme.Http, collectorConf.port, StatusCodes.MovedPermanently)
+          case _ => routes
+        }
```
istreeter, line 166: This `redirectTo` should only get called if the user enabled `collector.ssl.redirect` in the hocon.  Please can you check whether the config parameter is obeyed? I cannot see where the check is done.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
```diff
@@ -123,12 +124,15 @@ trait Collector {
...
-        extract(_.request.uri) { uri =>
-          redirect(
-            uri.copy(scheme = "https").withPort(collectorConf.ssl.port),
-            StatusCodes.MovedPermanently
-          )
-        }
+        redirectTo(Scheme.Https, collectorConf.ssl.port, StatusCodes.MovedPermanently)
+      }
+
+    def redirectTo(scheme: Scheme, port: Int, redirectionType: StatusCodes.Redirection) =
```
istreeter, line 130: I'm curious, why do you choose to pass `redirectionType` as a parameter if it can only ever take one value?

istreeter (changes requested):
@oguzhanunlu If you look at [the original issue](https://github.com/snowplow/stream-collector/issues/221) there are two different production setups that might need this feature.  Currently your implementation satisfies point 1 but we also need it for point 2.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
oguzhanunlu: I guess it is a premature optimization, thinking for future use where we might need different status code

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
oguzhanunlu: why do we need `collector.ssl.redirect` check for this? if `X-Forwarded-Proto: http` header exists and LB sent it to https port, I'd expect collector to respect the header and route the request to http port, regardless of `collector.ssl.redirect` setting as it is about redirecting http to https

istreeter: If `X-Forwarded-Proto: http` header exists, it means the tracker sent the request to the load balancer's http port.

If hocon has `collector.ssl.redirect = false` it means the pipeline operator is happy to accept tracker requests sent to the load balancer's http port.  So we should sink the request immediately and not issue a redirect.

If hocon has `collector.ssl.redirect = true` it means the pipeline operator does not want trackers to send requests to the load balancer's http port, for security reasons.  So the collector should return a `MovedPermanently` to make the tracker send the request to the load balancer's port.

To be clear.... the status code we return here is an instruction to the tracker on what to do next, it is not an instruction to the load balancer.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala
```diff
@@ -131,6 +131,20 @@ trait Collector {
         }
       }
 
+    def xForwardedProto(routes: Route): Route =
+      if (!collectorConf.ssl.redirect) routes
```
istreeter, line 135: The collectorConf.ssl also has a `enable` parameter.  I dislike this `enable` parameter immensely but for legacy reasons I wonder if we should respect it.

```
if (collectorConf.ssl.redirect.enable && collectorConf.ssl.redirect)
  optionalHeaderValueByName("X-Forwarded-Proto") { .... }
else routes
```

oguzhanunlu: good point, I'll respect that for now and we can remove that later

istreeter (approved):

PR #226, opened 2022-03-04
Replace metrics with statsd

oguzhanunlu:


core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -136,10 +136,15 @@ package model {
...
-    enabled: Boolean,
-    durationBucketsInSeconds: Option[List[Double]]
+
+  final case class StatsdConfig(
+    hostname: String,
+    port: Int,
+    period: FiniteDuration = 1.minute,
+    prefix: String         = "snowplow.collector"
   )
+  final case class MetricsConfig(enabled: Boolean, statsd: StatsdConfig)
```
istreeter, line 146: I wonder if `enabled` should be a sub-field of `statsd`.  Just in case we ever want a different type of metrics to enable/disable separately.

For example, many of our snowplow applications have a `metrics` block with both `stdout` and `statsd` options and they can be enabled independently e.g. [here in rdb loader](https://github.com/snowplow/snowplow-rdb-loader/blob/3.0.0/config/redshift.config.reference.hocon#L80-L98).

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -175,7 +180,7 @@ package model {
     rootResponse: RootResponseConfig,
     cors: CORSConfig,
     streams: StreamsConfig,
-    prometheusMetrics: PrometheusMetricsConfig,
+    metrics: Option[MetricsConfig],
```
istreeter, line 183: I think this does not need to be an `Option` and I think the code is simpler if it's not an option.

The `reference.conf` ensures that it always has a value at run time with default `enabled=false`, so from user's point of view it is optional.

istreeter (commented):
It looks good.  In other snowplow we nest the `metrics` block inside a `monitoring` block e.g. [here](https://github.com/snowplow/snowplow-rdb-loader/blob/3.0.0/config/redshift.config.reference.hocon#L55) and [here](https://github.com/snowplow/enrich/blob/3.1.2/config/config.kinesis.extended.hocon#L327).

But collector configuration is already completely inconsistent with all our other apps, so I'm not going to suggest we conform to what the other apps do.  At least not until a 3.0.0 release!

istreeter (approved):

PR #243, opened 2022-06-11
Release 2.6.1

istreeter:


kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -185,10 +185,10 @@ class KinesisSink private (
...
             )
             if (failurePairs.nonEmpty) {
-              failurePairs.foreach(f =>
-                log.error(
-                  s"Failed writing record to Kinesis stream $streamName, with error code [${f._2.getErrorCode}] and message [${f._2.getErrorMessage}]."
-                )
+              val errorCodes = failurePairs.map(_._2.getErrorCode).toSet.mkString(",")
+              val exampleMsg = failurePairs.map(_._2.getErrorMessage).find(_.nonEmpty).getOrElse("")
+              log.error(
+                s"Failed writing ${failurePairs.size} records to Kinesis stream $streamName, with error codes [$errorCodes] and example message [$exampleMsg]."
```
dilyand, line 191: It might be more helpful if we have an example message for each error code, but not a big deal if you'd rather go ahead as is.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -224,11 +224,11 @@ class KinesisSink private (
...
             if (s.nonEmpty) {
-              s.foreach { f =>
-                log.error(
-                  s"Failed writing message to SQS queue ${sqs.sqsBufferName}, with error code [${f._2.code}] and message [${f._2.message}]."
-                )
-              }
+              val errorCodes = s.map(_._2.code).toSet.mkString(",")
+              val exampleMsg = s.map(_._2.message).find(_.nonEmpty).getOrElse("")
+              log.error(
+                s"Failed writing ${s.size} records to SQS queue ${sqs.sqsBufferName}, with error codes [$errorCodes] and example message [$exampleMsg]."
```
dilyand, line 230: Same here.

dilyand (approved):
LGTM @istreeter . I suggested a slight modification to make the logs more helpful but feel free to log it as an issue to address in a future release (if you agree with it).

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
istreeter: You're completely right.  I just pushed an update.

PR #246, opened 2022-07-20
Send warmup requests to self on startup

istreeter:


core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Warmup.scala
```diff
@@ -0,0 +1,67 @@
...
+  def run(interface: String, port: Int, config: WarmupConfig)(
+    implicit ec: ExecutionContext,
+    system: ActorSystem
+  ): Future[Unit] =
+    if (config.enable) {
+      logger.info(s"Starting warm up of $interface:$port.  It is expected to see a few failures during warmup.")
+
+      val cxnSettings = ConnectionPoolSettings(system)
+        .withMaxConnections(config.maxConnections)
+        .withMaxOpenRequests(1 << 30) // must exceed maxConnections and must be a power of 2
```
benjben, line 43: We will probably never reach OOM even with high `numRequests` on a small JVM, but just for fun:
```suggestion
        .withMaxOpenRequests(Integer.highestOneBit(config.maxConnections - 1) * 4) // must exceed maxConnections and must be a power of 2
```
would give the *next* bigger power of 2 (e.g. if `maxConnextions` is 1024 then it will be 2048, so that it always exceeds).

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Warmup.scala
```diff
@@ -0,0 +1,67 @@
...
+        .withMaxRetries(0)
+
+      Source(1 to config.numRequests)
+        .map(_ => (HttpRequest(uri = s"/health"), ()))
+        .via(Http().cachedHostConnectionPool[Unit](interface, port, cxnSettings))
```
benjben, line 48: Is this automatically closed after the warm up?

benjben (approved):
Looks great :+1:

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Warmup.scala
istreeter: Very clever idea!  I think it fails if maxConnections is equal to 1.  But I think I can change it to `Integer.highestOneBit(config.maxConnections) * 2` and then get a sensible value for all numbers.

core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Warmup.scala
istreeter: Yes, cachedHostConnectionPool [returns a Flow](https://github.com/akka/akka-http/blob/v10.2.9/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala#L584) which is a akka-streams concept which manages its own state and closes itself when the Source terminates.

PR #250, opened 2022-08-16
RabbitMQ

benjben:
With RabbitMQ events are sent to what is called an exchange.

[This line](https://github.com/snowplow/stream-collector/blob/feature/RabbitMQ/rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala#L79) makes sure that the exchange exists when the app starts and returns an exception if not.

I've tested that when the connection to RabbitMQ gets lost, events are sent when the connection is back.

### DIFF
--- a/rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
+++ b/rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2013-2022 Snowplow Analytics Ltd. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package com.snowplowanalytics.snowplow.collectors.scalastream
+package sinks
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+
+import cats.syntax.either._
+
+import com.rabbitmq.client.Channel
+
+import com.snowplowanalytics.snowplow.collectors.scalastream.model.RabbitMQBackoffPolicyConfig
+
+class RabbitMQSink(
+  channel: Channel,
+  exchangeName: String,
+  backoffPolicy: RabbitMQBackoffPolicyConfig,
+  executionContext: ExecutionContext
+) extends Sink {
+
+  implicit val ec = executionContext
+
+  override val MaxBytes = 128000000
+
+  override def storeRawEvents(events: List[Array[Byte]], key: String): Unit =
+    if (events.nonEmpty) {
+      log.info(
+        s"Sending ${events.size} Thrift records to exchange $exchangeName"
+      )
+      Future.sequence(events.map(e => sendOneEvent(e))).onComplete {
+        case Success(_) =>
+          log.debug(
+            s"${events.size} events successfully sent to exchange $exchangeName"
+          )
+        // We should never reach this as the writing of each individual event is retried forever
+        case Failure(e) =>
+          throw new RuntimeException(s"Error happened during the sending of ${events.size} events: ${e.getMessage}")
+      }
+    }
+
+  private def sendOneEvent(bytes: Array[Byte], currentBackOff: Option[FiniteDuration] = None): Future[Unit] =
+    Future {
+      if (currentBackOff.isDefined) Thread.sleep(currentBackOff.get.toMillis)
+      channel.basicPublish(exchangeName, "", null, bytes)
+    }.recoverWith {
+      case e =>
+        val nextBackOff =
+          currentBackOff match {
+            case Some(current) =>
+              (backoffPolicy.multiplier * current.toMillis).toLong.min(backoffPolicy.maxBackoff).millis
+            case None =>
+              backoffPolicy.minBackoff.millis
+          }
+        log.error(s"Sending of event failed with error: ${e.getMessage}. Retrying in $nextBackOff")
+        sendOneEvent(bytes, Some(nextBackOff))
+    }
+
+  override def shutdown(): Unit = ()
+}
+
+object RabbitMQSink {
+  def init(
+    channel: Channel,
+    exchangeName: String,
+    backoffPolicy: RabbitMQBackoffPolicyConfig,
+    executionContext: ExecutionContext
+  ): Either[Throwable, RabbitMQSink] =
+    for {
+      _ <- Either.catchNonFatal(channel.exchangeDeclarePassive(exchangeName))
+    } yield new RabbitMQSink(channel, exchangeName, backoffPolicy, executionContext)
+}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,6 +23,7 @@ jobs:
           sbt 'project pubsub' assembly
           sbt 'project sqs' assembly
           sbt 'project stdout' assembly
+          sbt 'project rabbitmq' assembly
       - name: Get current version
         id: ver
         run: |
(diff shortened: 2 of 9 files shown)


snowplowcla:
Thanks for your pull request. Is this your first contribution to a Snowplow open source project? Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

:memo: Please visit https://docs.snowplowanalytics.com/docs/contributing/contributor-license-agreement/ to learn more and sign.

Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
```diff
@@ -0,0 +1,81 @@
...
+
+import com.snowplowanalytics.snowplow.collectors.scalastream.model.RabbitMQBackoffPolicyConfig
+
+class RabbitMQSink(
+  channel: Channel,
+  exchangeName: String,
+  backoffPolicy: RabbitMQBackoffPolicyConfig
+) extends Sink {
+
+  implicit val ec = ExecutionContext.global
```
peel, line 32: Shouldn't we be running sink in its own thread pool?
Given there is a slowdown on AMQP messaging it might clog the main thread pool making the collector unresponsive. 
I believe we should stick it into its own (probably fixed size) thread pool to avoid it interfering with HTTP layer.

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
```diff
@@ -0,0 +1,81 @@
...
+            s"${events.size} events successfully sent to exchange $exchangeName"
+          )
+        // We should never reach this as the writing of each individual event is retried forever
+        case Failure(e) =>
+          throw new RuntimeException(s"Error happened during the sending of ${events.size} events: ${e.getMessage}")
+      }
+    }
+
+  private def sendOneEvent(bytes: Array[Byte], currentBackOff: Option[FiniteDuration] = None): Future[Unit] =
+    Future {
```
peel, line 53: I think this is where it manifests the most. If I recall correctly this will be running on main execution context's thread pool (not global, but akka actor system's top-level). The Future will block the thread it's executed on. Might not be of a huge impact but once AMQP goes down this might saturate the top-level thread pool with retries.
Another way to do this would be to make use of `akka.ActorSystem`'s  `scheduler.schedule*` method.

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
```diff
@@ -0,0 +1,81 @@
...
+      if (currentBackOff.isDefined) Thread.sleep(currentBackOff.get.toMillis)
+      channel.basicPublish(exchangeName, "", null, bytes)
```
peel, line 55: Would it be worth to be able to allow setting routing keys? This allows you to build pretty impressive topologies. 
This one in particular would allow us to deploy canary-builds. Setting routing key to `latest` on 10% of collector instances when the rest of the instances are `stable` allow us to have a deployment-check for consumer for the `unstable`. If succeeds we can roll out 100%.
This would also allow having an interesting topology where parts of the messages are delivered to both enrich and another destination. 
Anyway, I feel like it might be worth it to expose it as a config parameter.

peel (commented):
Looking great @benjben! I think it's simple yet very important development.
I've added a couple performance-related concerns and it would be great if you considered these.
Another thing I'd propose is setting up an integration test using test containers rabbitmq (or docker) modules. We'd be able to run an end-to-end rabbit collector test locally.

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
benjben: Very good point! I had actually thought of doing that, and didn't do it because I didn't know how to decide the size of this thread pool (I'm ashamed). What would you recommend?

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
benjben: First it was exposed and then I removed it because when we were testing Josh said that we would not use it but I agree that it doesn't hurt to expose it anyway.

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
benjben: I considered using `akka.ActorSystem`'s `scheduler.schedule*` but at the time when we create the sink, `ActorSystem` does not exists yet, and it would require some refactor in the design of the collector, so I didn't go this way

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
peel: Nothing to be ashamed of! This is a really tricky topic and the consequences are unobvious. It is also very much dependent on runtime environment so I believe it should be configurable by the user. And optional. As a default, in terms of how we determine the thread pool size I think [this](https://engineering.zalando.com/posts/2019/04/how-to-set-an-ideal-thread-pool-size.html) summarises in pretty well. I've followed the Goetz approach as a starting point for when I had to set these things up before. And it usually works for usual cases. But I think it's a perf tuning thing, and just sticking it into another threadpool of some size will make things much more resilient. Perf improvements probably come in later especially as we're introducing entirely new transport layer.

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
peel: I've spent some quality time using and abusing rabbit before and the ability to use various topologies is what is really exciting in this PR. I think having that option might come in handy to some OSS users who'd like to use their existing rabbit setup. Especially as this partially implements multiple purpose pipelines people have been asking before. 
I think part of the announcement, apart from the cloud-agnosticism, could be "all the cool things you could do with the new transport". And topologies are really interesting topic.

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
peel: Nothing to be ashamed of! This is a tricky one. Especially as it's very much runtime-dependent. I'd expose an optional setting for this. Now, about how we set the default it probably the interesting part. I think [this](https://engineering.zalando.com/posts/2019/04/how-to-set-an-ideal-thread-pool-size.html) is a pretty useful summary.
All in all I've usually followed the Goetz's approach and as a starting point it served well. Now, as we're introducing a new transport layer we will do lots of performance testing and tuning along the way. I'd start with some rough calculation using the `Runtime.getRuntime().availableProcessors()` and we'll be able to override that if necessary.

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
peel: I've spent some quality time using and abusing rabbit in my previous life. And part of my excitement for this new transport layer is being able to set up those topologies. Apart from the cloud-agnosticism, it feels like a "feature" we should be able to use as a part of the announcement. "All the cool topologies you could build with the new collector/enrich".

rabbitmq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/RabbitMQSink.scala
benjben: About `Runtime.getRuntime().availableProcessors()`:
```
But be careful with this command if you run your service in a container environment. Without specifying any constraints, a containerized process will be able to see the hardware on the host OS.
```

I find it super hard to decide a default size, especially because it depends mostly about RabbitMQ and collector can't know that. So here is what I propose: make thread pool size configurable and optional. If it is set, use it for the fixed pool size. If it's not there, use a cached thread pool.

In both cases, we can't avoid an OOM if RabbitMQ goes down because the collector will keep accepting new requests. In both cases, RabbitMQ won't get flooded with requests when it's up again because all the requests go through the same channel.

WDYT?

peel: > About Runtime.getRuntime().availableProcessors():

I think it's been like this prior to JDK11. It introduced `-XX-UseContainerSupport` (as a non-default) which fixed a bunch of resource issues for containers. JDK17 fixes even more. But yes, totally depends on the runtime *we pick*. We had to set flags on containers and even then still, java lied about what's available to the container because it wasn't aware of the cgroups (v2!) settings.

So, the behaviour is: if you run java on a host that only supports groups v2 and not groups v1, you will get the host's resources. Meaning, we need to consider that.

> make thread pool size configurable and optional. If it is set, use it for the fixed pool size. If it's not there, use a cached thread pool.
Makes sense. We totally need to document the runtime and consider the JDK platform we use for build/runtime.

benjben:
Hey @peel , I addressed your feedback and made the thread pool size optional, as we discussed, please have a look.

Regarding the integration tests, we will add it in the testing squad when we will have agreed on a methodology.

PR #255, opened 2022-09-02
Kafka integration tests, first iteration

TiganeteaRobert:
Added two integration tests for the Kafka collector:

- Test that we are not losing events. Sending 55 events in a row to the collector and counting the number of events in the `good` Kafka topic.
- Test that an oversized event gets to the `bad` topic.


.github/workflows/kafka-test.yml
```diff
@@ -0,0 +1,64 @@
...
+          (cd integration && docker-compose -f ./docker-compose.yml up -d)
+          sleep 10
+      - name: Run collector
+        run: |
+          docker run --network="integration_net1" --name collector -d -v "$PWD"/.github/workflows/kafka-collector-config:/snowplow/config -p 12345:12345 snowplow/scala-stream-collector-kafka:0.0.0 --config /snowplow/config/config_good.hocon
+          sleep 15
+      - name: Send good events
+        run: |
+          while read p; do
+            curl -X POST -i http://0.0.0.0:12345/com.snowplowanalytics.snowplow/tp2  -d "$p" 2>&1 | grep -q 'Set-Cookie'
```
istreeter, line 29: What is the purpose of `grep -q 'Set-Cookie'` for this test?

.github/workflows/kafka-collector-config/config_good.hocon
```diff
@@ -0,0 +1,85 @@
+collector {
```
istreeter, line 1: There's nothing wrong with this config, but he new minimal configs are allowed to be much shorter, see https://github.com/snowplow/stream-collector/blob/2.7.0/examples/config.kafka.minimal.hocon

.github/workflows/kafka-test.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Create integration resources
+        run: |
+          (cd integration && docker-compose -f ./docker-compose.yml up -d)
+          sleep 10
+      - name: Run collector
+        run: |
+          docker run --network="integration_net1" --name collector -d -v "$PWD"/.github/workflows/kafka-collector-config:/snowplow/config -p 12345:12345 snowplow/scala-stream-collector-kafka:0.0.0 --config /snowplow/config/config_good.hocon
+          sleep 15
+      - name: Send oversized event to collector
+        run: curl -X POST -i http://0.0.0.0:12345/com.snowplowanalytics.snowplow/tp2  -d @"$PWD"/.github/workflows/integration_tests/kafka/1mb_string 2>&1 | grep -q 'Set-Cookie'
```
istreeter, line 60: I think it would be very neat if we could dynamically generate the 1 MB string using bash, instead of committing it to the repo.

istreeter (commented):
This looks great @TiganeteaRobert well done!

I have a slight concern about all the sleeps and timeouts -- I wonder how resilient this test will be when it is run by a github runner.  It will be annoying if a timeout on an integration test prevents us from successfully tagging a release.

Instead of sleeping for a fixed time, maybe we need a bash function that waits for a port to be open.  Something like [this](https://unix.stackexchange.com/a/5279).  It's a problem that will come up again and again as we write more integration tests, so it's worth finding a good solution now.

.github/workflows/kafka-test.yml
TiganeteaRobert: I've thought about that too, let me see what I can do!

.github/workflows/kafka-test.yml
TiganeteaRobert: It's a copy-paste leftover from the cookieless workflow from which I've taken the curl command, where we check for this value. Removing it.

.github/workflows/kafka-collector-config/config_good.hocon
TiganeteaRobert: Sure, I'll reduce the config as much as possible.

.github/workflows/kafka-test.yml
```diff
@@ -0,0 +1,64 @@
...
+          java-version: 11
+      - name: Publish Docker image
+        run: sbt 'project kafka; set Docker / version := "0.0.0"' docker:publishLocal
+      - name: Create integration resources
+        run: |
+          (cd integration && docker-compose -f ./docker-compose.yml up -d)
+          sleep 10
+      - name: Run collector
+        run: |
+          docker run --network="integration_net1" --name collector -d -v "$PWD"/.github/workflows/kafka-collector-config:/snowplow/config -p 12345:12345 snowplow/scala-stream-collector-kafka:0.0.0 --config /snowplow/config/config_good.hocon
```
benjben, line 57: Is there any reason why not starting the collector with docker compose, along with zookeeper and Kafka?

.gitignore
```diff
@@ -1,7 +1,8 @@
 project/target
 project/project
 target
-
+.idea
 src/main/resources/application.conf
 
 .bsp
+.DS_Store
```
benjben, line 8: These files are not part of the project but of your local setup, they should be part of your own global gitignore, not this one

TiganeteaRobert: Agreed.

.github/workflows/kafka-test.yml
```diff
@@ -0,0 +1,64 @@
...
+          sleep 15
+      - name: Send oversized event to collector
+        run: curl -X POST -i http://0.0.0.0:12345/com.snowplowanalytics.snowplow/tp2  -d @"$PWD"/.github/workflows/integration_tests/kafka/1mb_string 2>&1 | grep -q 'Set-Cookie'
+      - name: Consume bad event from Kafka
+        run: timeout 5 docker exec -w /bin broker kafka-console-consumer.sh --topic=bad --from-beginning --bootstrap-server 0.0.0.0:9092 | grep -q "oversized"
```
benjben: Here we should count the bad events and fail if the count is different from expected

.github/workflows/kafka-test.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          (cd integration && docker-compose -f ./docker-compose.yml up -d)
+          sleep 10
+      - name: Run collector
+        run: |
+          docker run --network="integration_net1" --name collector -d -v "$PWD"/.github/workflows/kafka-collector-config:/snowplow/config -p 12345:12345 snowplow/scala-stream-collector-kafka:0.0.0 --config /snowplow/config/config_good.hocon
+          sleep 15
+      - name: Send oversized event to collector
+        run: curl -X POST -i http://0.0.0.0:12345/com.snowplowanalytics.snowplow/tp2  -d @"$PWD"/.github/workflows/integration_tests/kafka/1mb_string 2>&1 | grep -q 'Set-Cookie'
+      - name: Consume bad event from Kafka
```
benjben, line 61: It would be more convenient to have all the steps of the actual test (send events, consume output) in one bash script (e.g. `test-good.sh`) that takes config as environment variables (e.g. `COLLECTOR_URI`, `BOOTSTRAP_SERVERS`..). This way it becomes possible to run the test locally without needing to push to Github.

.github/workflows/kafka-test.yml
TiganeteaRobert: Yes, it seems that Kafka container does not properly register that it's been started (as in, the container is started, but Kafka is not ready). From my testing, even using `depends_on` in the docker-compose.yml file it does not wait properly for it to be ready.

.github/workflows/kafka-test.yml
TiganeteaRobert: Yes I just fixed that, also now generating a 1 MB string instead of getting it from a file.

.github/workflows/kafka-test.yml
TiganeteaRobert: I have opted towards defining steps in the GH workflow instead of creating a big bash script because the output of the separate steps is handles by GH automatically and it's much nicer to see right out where the test failed. I am flexible on this matter though, it's just my proposal.

TiganeteaRobert: But yeah, I agree that this makes running the integration tests locally pretty much impossible (there are tools such as `act` that emulate GH actions locally but they're kinda finicky and annoying to use from my experience)

.github/workflows/kafka-test.yml
TiganeteaRobert: Right now I'm trying to implement a check on the port as @istreeter proposed above to also get rid of the hardcoded 10 second sleep.

TiganeteaRobert: Added a check for existing topic instead.

.github/workflows/kafka-test.yml
TiganeteaRobert: Now generating a string dynamically!

.github/workflows/kafka-collector-config/config_good.hocon
TiganeteaRobert: Reduced it to the bare minimum.

benjben (commented):
To stay aligned with our strategy to decouple the creation of resources and running the app with the execution of the tests, here is what I would do:

- Add the collector to Docker compose, so that when we run `docker-compose up`, everything is running and ready for the tests. I know that the difficulty is that the collector would crash because it would start before Kafka topics are created. To solve that I would add another service in Docker compose (also using `kafka` image and `kafka-topics.sh`) that just waits until the topics are created (e.g. by using [until](https://linuxize.com/post/bash-until-loop/)). Then when it exists the collector can start successfully.

- Put the actual integration tests inside their bash scripts, that receive the endpoints through environment variables. I agree that the output won't be as nice as having different steps in GA UI, but the most important is that we have the possibility to know what went wrong to troubleshoot, and we will. And this way:
  1. We can run the tests locally
  2. It will be very easy to update the day we want to run the same test on a collector deployed with our infra (well, maybe not for Kafka but for Kinesis and PubSub). All we'll need to do will be to update the environment variables

What do you think of this plan?

.github/workflows/kafka-test.yml
```diff
@@ -0,0 +1,70 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: coursier/cache-action@v3
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Publish Docker image
+        run: sbt 'project kafka; set Docker / version := "0.0.0"' docker:publishLocal
```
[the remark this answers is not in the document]
TiganeteaRobert: The issue is that for each job GH spins up a different VM, I'll need to see how I can make it so that this command can only be ran once.

.github/workflows/kafka-test.yml
```diff
@@ -0,0 +1,70 @@
...
+          while read p; do
+            curl -X POST -i http://0.0.0.0:12345/com.snowplowanalytics.snowplow/tp2 -d "$p"
+          done <"$PWD"/.github/workflows/integration_tests/data.txt
+        shell: bash
+      - name: Wait for events to be sent from the collector to Kafka
+        run: sleep 5
+      - name: Check number of events in the good topic
+        run: docker exec -w /bin broker kafka-run-class.sh kafka.tools.GetOffsetShell --broker-list 0.0.0.0:9092 --topic good | grep 'good:0:55'
+      - name: Check no events got to the bad topic
+        run: docker exec -w /bin broker kafka-run-class.sh kafka.tools.GetOffsetShell --broker-list 0.0.0.0:9092 --topic bad | grep 'bad:0:0'
```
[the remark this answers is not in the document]
TiganeteaRobert: Yes it does, it exits with an error in the GH UI.

.github/workflows/kafka-test.yml
```diff
@@ -0,0 +1,70 @@
+name: Integration tests - Kafka
```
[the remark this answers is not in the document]
oguzhanunlu: GitHub recently [announced](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/) `workflow_run`, which lets you make a workflow depend on another workflow. Note that it doesn't care about the successful completion. See [this](https://github.com/orgs/community/discussions/26238#discussioncomment-3250901) to trigger your downstream workflow on the success of upstream.

.github/workflows/kafka-test.yml
```diff
@@ -0,0 +1,70 @@
...
+      - uses: coursier/cache-action@v3
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Publish Docker image
+        run: sbt 'project kafka; set Docker / version := "0.0.0"' docker:publishLocal
+      - name: Create integration resources
+        run: |
+          (cd integration && docker-compose -f ./docker-compose.yml up -d)
```
oguzhanunlu, line 20:
```suggestion
          docker-compose -f integration/docker-compose.yml up -d
```

.github/workflows/kafka-test.yml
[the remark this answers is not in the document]
oguzhanunlu: not sure if you already found it but let me share [this page](https://github.com/docker/build-push-action/blob/master/docs/advanced/share-image-jobs.md) which explains how to share an image across jobs

.github/workflows/kafka-test.yml
oguzhanunlu: Kafka offer this mechanism via [`reconnect.backoff.max.ms`](https://kafka.apache.org/32/documentation/#producerconfigs_reconnect.backoff.max.ms) , it is 1 sec by default but we could set it to e.g. 10 sec and let it retry failed broker connection attempts up to 10 sec.

oguzhanunlu: if we can bump kafka-clients to 2.7 or later, we can benefit from [`socket.connection.setup.timeout.max.ms`](https://kafka.apache.org/32/documentation/#producerconfigs_socket.connection.setup.timeout.max.ms) config which could let us avoid current retry impl and let kafka-clients do it

that could let us run all in docker-compose as well

integration/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    image: wurstmeister/zookeeper
+    container_name: zookeeper
+    ports:
+      - "2181:2181"
+    restart: unless-stopped
+    networks:
+      - net1
+
+  broker:
+    image: wurstmeister/kafka:latest
```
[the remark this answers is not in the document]
oguzhanunlu: assuming that the reason is to create topics automatically upon the launch of broker, we can benefit from `auto.create.topics.enable` config of Kafka and it is true by default in [Kafka 2.2.x](https://kafka.apache.org/22/documentation.html)

PR #257, opened 2022-09-13
Kinesis integration test, working

TiganeteaRobert:


.github/workflows/integration_tests/no-data-loss/consume_kinesis.sh
```diff
@@ -1,22 +1,31 @@
...
+lines=$(aws --endpoint-url=http://localhost:4566 kinesis get-records --shard-iterator $iterator --output text --region eu-central-1 | tail -n +2|awk '{print $3}' & sleep 1; kill $!)
+
+for l in $lines
   do
-    echo $l|base64 -d
-    echo -e "\n"
-  done
-  if [[ -n $lines ]]
+  found=false
+  while read p; do
```
TiganeteaRobert, line 17: @dilyand I've also reversed the logic here to the correct one.

.github/workflows/kinesis-test.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: coursier/cache-action@v3
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Initialize the resources and the collector and send data to the collector
+        run: '"$PWD"/integration/init.sh kinesis "$PWD"/integration/docker-compose.yml integration.hocon "$PWD"/.github/workflows/integration_tests/no-data-loss/data.txt'
```
TiganeteaRobert, line 16: Here you can see how the initialisation script works. This allows us to locally spin up the required resources so the tests can be ran locally.

.github/workflows/kinesis-test.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Read from Kinesis good stream
+        run: ./.github/workflows/integration_tests/no-data-loss/consume_kinesis.sh good "$PWD"/.github/workflows/integration_tests/no-data-loss/data.txt
```
TiganeteaRobert, line 18: This script can be ran locally after the resources are started.

TiganeteaRobert:
Right now if we want to run the tests locally, we have to run the resource init script, then the consume script. Should we create a script that runs these two in sequence or is that overkill?

Makefile
```diff
@@ -0,0 +1,19 @@
...
+
+kinesis-no-data-loss-consume: kinesis-no-data-loss-integration-up
+	./.github/workflows/integration_tests/no-data-loss/consume_kinesis.sh good ./.github/workflows/integration_tests/no-data-loss/data.txt
+
+kinesis-no-data-loss: kinesis-no-data-loss-integration-up kinesis-no-data-loss-consume
+
+kinesis-integration-tests: kinesis-no-data-loss-consume
+# ----------
+
+integration-tests: kinesis-integration-tests integration-down
```
TiganeteaRobert, line 19: This `Makefile` allows the user to run the integration tests locally.

.github/workflows/integration_tests/no-data-loss/consume_kinesis.sh
```diff
@@ -1,19 +1,21 @@
...
 
-shard=$(aws --endpoint-url=http://localhost:4566 kinesis describe-stream --stream-name $streamName --output text | grep SHARDS | awk '{print $2}')
-iterator=$(aws --endpoint-url=http://localhost:4566 kinesis get-shard-iterator --stream-name $streamName --shard-id $shard --shard-iterator-type TRIM_HORIZON --output text)
-lines=$(aws --endpoint-url=http://localhost:4566 kinesis get-records --shard-iterator $iterator --output text | tail -n +2|awk '{print $3}' & sleep 1; kill $!)
+shard=$(aws --endpoint-url=http://localhost:4566 kinesis describe-stream --stream-name $streamName --output text --region eu-central-1 | grep SHARDS | awk '{print $2}')
+
+iterator=$(aws --endpoint-url=http://localhost:4566 kinesis get-shard-iterator --stream-name $streamName --shard-id $shard --shard-iterator-type TRIM_HORIZON --output text --region eu-central-1)
+lines=$(aws --endpoint-url=http://localhost:4566 kinesis get-records --shard-iterator $iterator --output text --region eu-central-1 | tail -n +2 | awk '{print $3}' & sleep 1 ; kill $!)
 
 while read p; do
```
TiganeteaRobert, line 14: I had to revert back the logic of this test. It now looks in the data.txt file, and looks for each line in Kinesis.

The reason for this is that the output of `get-records` is really hard to parse and extract the actual data that was sent to the collector, so this makes it easier for further tests to be implemented.

IMO this logic should be fine too, but if it's needed I'll find a way to use the reverse logic.

dilyand:
Hey @TiganeteaRobert , I tried running these tests locally but couldn't.

First I tried:

```
cd snowplow/stream-collector
make
```

but this only runs the `integration-down` step of the Makefile, as it comes first (does the Makefile need re-ordering?)

Then I tried to run the `integration-tests` step:

```
make integration-tests
```

which failed.

I tried to run each step of the `integration-tests` target separately, like it's done in the `kinesis-test.yml` file:

```
make kinesis-no-data-loss-integration-up
```

and got this error:

> Error response from daemon: invalid mount config for type "volume": invalid mount path: 'localstack/tmp' mount path must be absolute

which I believe refers to [this line](https://github.com/snowplow/stream-collector/blob/6a405ec450677acbbb06d770f9faeecab491313f/integration/docker-compose.yml#L10) in `docker-compose.yml`. It seems like this step has succeeded in the GH action that was triggered when you pushed your changes. Any idea why it works there but not locally?

Anyway, the whole step ultimately only created the collector container, but no healthy localstack instance.

So, when I ran:

```
make kinesis-no-data-loss-send-data
```

it worked, since the collector was running.

But then:

```
make kinesis-no-data-loss-consume
```

fails with:

> Could not connect to the endpoint URL: "http://localhost:4566/"

because localstack is not running.

I observed the same when running:

```
make kinesis-oversized
```

PR #259, opened 2022-09-27
Make collector timestamp timezone configurable

hamnis:
A customer wants to control the partition in BigQuery to avoid having to query multiple partitions. To make this possible, we want to control which timezone the current time millis are collected.

This should keep the default to UTC as System.currentTimeMillis is, but allow collectors that are in different timezones to be able to use a different timezone if needed.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -94,3 +94,5 @@ experimental {
     maxConnections = 2000
   }
 }
+
+timezone = "UTC"
--- a/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
+++ b/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
@@ -16,19 +16,17 @@ package com.snowplowanalytics.snowplow.collectors.scalastream
 
 import java.net.{MalformedURLException, URL}
 import java.nio.charset.StandardCharsets.UTF_8
-import java.time.Instant
+import java.time.{Instant, ZoneId, ZoneOffset, ZonedDateTime}
 import java.util.UUID
 import org.apache.commons.codec.binary.Base64
 import org.slf4j.LoggerFactory
 
 import scala.collection.JavaConverters._
-
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.model.headers.CacheDirectives._
 import cats.data.NonEmptyList
 import cats.implicits._
-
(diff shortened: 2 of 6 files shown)


istreeter:
Hi @hamnis thank you for taking the time to submit a change to this project 🙌

Currently, the `timestamp` field emitted by the collector is a standard "unix" timestamp which has a strict definition: The number of milliseconds since 00:00:00 Thursday, 1 January 1970 **in the UTC timezone**.  By changing the timezone here in the collector, it is twisting the meaning of this timestamp to be something very non-standard.

Snowplow has a lot of downstream tooling which relies on the collector's timestamp to have the standard meaning.  So I don't think we can change the timestamp here in the collector, because downstream tooling would then incorrectly read the timestamp as the wrong time.

I understand that you want to control the partitions in BigQuery so you can make more efficient day-based queries -- that's an interesting requirement.  I think it is better achieved by looking at how we load Snowplow events into BigQuery.

Currently in the Snowplow BigQuery events table we set all timestamp columns as the `Timestamp` type.  This type again has a strict fixed meaning regarding timezone, [described here](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp_type):

> A TIMESTAMP value represents an absolute point in time, independent of any time zone or convention such as Daylight Savings Time, with microsecond precision.

[And this page](https://cloud.google.com/bigquery/docs/partitioned-tables#date_timestamp_partitioned_tables) states how BigQuery partitions tables using a Timestamp column:

> Partitions boundaries are based on UTC time.

So to facilitate what you want, I think you need a column in the events table to have a `Datetime` type [described over here](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#datetime_type) instead of Timestamp.  And you would need Snowplow to load the column using a configurable timezone.  Currently this is not possible with our [Snowplow Bigquery loader](https://github.com/snowplow-incubator/snowplow-bigquery-loader), so it would be a new feature request to implement it over there.

It would be interesting to hear more details about what your queries you run, and why this partitioning scheme would help.  We're probably more likely to implement something new like this if we understand better how it enables new use cases in the warehouse.

hamnis:
I understand that this might not be the right place to do this. My initial thought was to do time conversion as soon as possible, as the collector timestamp is the partition time stored in BQ. 

Currently in our setup the raw events table is partitioned using DAYs. 

To do aggregations on full days, in our timezone, we would then need to might need to query 3 partitions: Previous day, current day and next day to get the full 24 hours. This is highly inefficient and we would like to be able to query the full day in one partition. 

I can do the changes in the correct repository.

PR #261, opened 2022-09-30
Prune system libraries from distroless image

istreeter:


build.sbt
```diff
@@ -85,7 +85,61 @@ lazy val dockerSettingsDistroless = Seq(
...
+        "/lib/x86_64-linux-gnu/libcrypt.so.*",
+        "/lib/x86_64-linux-gnu/libexpat.so*",
+        "/lib/x86_64-linux-gnu/libmemusage.so*",
+        "/lib/x86_64-linux-gnu/libmvec-*.so*",
+        "/lib/x86_64-linux-gnu/libnsl-*.so*",
+        "/lib/x86_64-linux-gnu/libnss_hesiod-*.so*",
+        "/lib/x86_64-linux-gnu/libpcprofile.so*",
+        "/lib/x86_64-linux-gnu/libpcre.so*",
+        "/lib/x86_64-linux-gnu/libutil-*.so*",
+        "/lib/x86_64-linux-gnu/libthread_db-*.s"
```
benjben, line 139: Maintaining this will be such a pain. I suggest that:
1. We leave the default configuration untouched (this image will never get published to DockerHub, only locally)
2. We automate what you've done in a script that will:
  a. Publish the Docker image locally
  b. Run it
  c. Get the libraries that are used by looking at `/proc`
  d. Prepare the `Dockerfile` by using the result from `c`
  e. Publish the image

istreeter: Thanks Ben, you were right it would be a pain to maintain it as it was.  I was thinking it but I needed someone else to say it!

> We leave the default configuration untouched (this image will never get published to DockerHub, only locally)

I would like it so the docker build is a sbt task, rather than a making script that runs completely independently of sbt.  It's because of the integration tests we are working on.  We want it so that `sbt test` builds and tests exactly the image that will be used in production.  Instead of testing a "local" version which is different to what gets published to Docker hub.

But... inspired by your suggestion, I refactored so there is now a install script.  Unfortunately it does not run java and look at `/proc` like you suggested -- because I couldn't get this working.  But even so I think it looks more easily maintainable than it did before.

.github/workflows/deploy.yml
```diff
@@ -130,10 +130,10 @@ jobs:
           push: true
 
       - name: Push distroless image
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v3
         with:
           context: distroless/${{ matrix.platform }}/target/docker/stage
           file: distroless/${{ matrix.platform }}/target/docker/stage/Dockerfile
-          platforms: linux/amd64,linux/arm64/v8
```
benjben, line 137: So we drop arm support ?

build.sbt
```diff
@@ -81,11 +81,25 @@ lazy val dockerSettingsDistroless = Seq(
...
+  dockerPermissionStrategy := DockerPermissionStrategy.CopyChown,
+
+  Docker / mappings := (Docker / mappings).value ++ Seq(new File("docker/install.sh") -> "install.sh"),
+
+  Docker / dockerCommands := {
+    Seq(
+      Cmd("FROM", "debian:bullseye-slim", "AS", "bullseye"), // Provides standard linux executables for the install script
+      Cmd("COPY", "install.sh", "/usr/bin/"),
+      Cmd("FROM", dockerBaseImage.value, "AS", "java"), // Provides a java installation
+      Cmd("FROM", "gcr.io/distroless/base:nonroot"), // The true base image
```
benjben, line 97: I'm surprised that we can have several `FROM` like this. What does this do exactly ?

docker/install.sh
```diff
@@ -0,0 +1,18 @@
...
+cp /java/lib/x86_64-linux-gnu/libz.so* /java/lib/x86_64-linux-gnu/libgcc_s.so* /lib/x86_64-linux-gnu/
+cp /java/usr/lib/x86_64-linux-gnu/libstdc++.so* /usr/lib/x86_64-linux-gnu/
+
+# Install java
+cp -r /java/etc/java-11-openjdk /etc/
+cp -r /java/etc/ssl/certs/java /etc/ssl/certs/
+cp -r /java/usr/lib/jvm /usr/lib/
+
+# Remove libs and executables from the base image we don't want
+rm /usr/lib/x86_64-linux-gnu/libssl.so*
```
benjben, line 17: What if we add a Scala dependency that is based on this lib ?

benjben (commented):
We still need to manually maintain the list of `.so` that we need and don't need, right ?

I'm not entirely sure to get the new logic with the building of the Docker image, but would this be feasible : in the `build.sbt` we have 2 different Docker settings, one for e.g. `collector-kinesis-tmp` that is the same as before, and the true `collector-kinesis` that gets published, and for the latter we have an install script like you did that runs `collector-kinesis-tmp`, looks at the libraries used, and deletes the other ones on `collector-kinesis` ? This way we could still use this Docker image in the integration tests and the listing of libraries would be fully automated.

build.sbt
benjben: > It's because of the integration tests we are working on. We want it so that sbt test builds and tests exactly the image that will be used in production

Very good point !

.github/workflows/deploy.yml
istreeter: Whoops, I deleted that by accident (just while testing).  I'll put it back again.

build.sbt
istreeter: This documentation describes it better than I can! https://docs.docker.com/build/building/multi-stage/

docker/install.sh
istreeter: Do you know of any scala or java dependency that depends on this lib?  Most of the time, scala/java dependencies make no assumptions about what system libs are installed on the host.  There are a few exceptions which depend directly on system libraries (e.g. the lzo serializer we use in s3 loader) but they are rare.

And furthermore... if we do ever need it then we amend this line.  It will be a new requirement of our base image, so we adjust the libraries to meet the new requirement.

istreeter:
> We still need to manually maintain the list of .so that we need and don't need, right ?

Yes.  The list of .so that we _don't_ want is now two items long: libssl and openssl.  That feels maintainable.  It's just saying we don't need to use native ssl directly.

The list of .so that we _do_ need is a bit longer.  I admit this is where we have to call the balance about whether it's worth it.

My reasons for being comfortable with this are:

- The libraries I have listed are very standard; nothing too exotic.
- We are pinned to a java version (11) so it is extremely unlikely the system requirements will change before we bump the java version.
- If we ever get the list wrong... then integration tests will fail.  I say this as a positive thing; we get notification that we need to fix something.

> I'm not entirely sure to get the new logic with the building of the Docker image, but would this be feasible : in the build.sbt we have 2 different Docker settings, one for e.g. collector-kinesis-tmp that is the same as before, and the true collector-kinesis that gets published, and for the latter we have an install script like you did that runs collector-kinesis-tmp, looks at the libraries used, and deletes the other ones on collector-kinesis ? This way we could still use this Docker image in the integration tests and the listing of libraries would be fully automated.

It's probably technically possible but very difficult to implement.  But my bigger reason for not doing with a script.... I don't think I would trust the result!  The script that you envisage would remove every .so that is not being linked when measured at a single point in time.  But I would not trust that the .so might be needed some time later.  I have more trust that the curated list of known libs are the ones we do/don't need.

Are you willing to prototype the script you're describing?  Very sadly I don't think I have time for it :(

build.sbt
benjben: Interesting, thanks for sharing!

docker/install.sh
benjben: > Do you know of any scala or java dependency that depends on this lib?

I don't, and honestly it's quite obscure to me how to know if a certain library that we use requires a `.so` under the hood or not.

benjben:
> The script that you envisage would remove every .so that is not being linked when measured at a single point in time. But I would not trust that the .so might be needed some time later.

I was wondering exactly that! Then there is a risk that integration tests don't cover an edge case where a `.so` would be used. But that's probably acceptable, and we can add this edge case to the tests when we know about it.

> Are you willing to prototype the script you're describing?

Because of the reason from above, that's probably not worth it.

> My reasons for being comfortable

I'm honestly not very comfortable with maintaining this `install.sh` and these custom Docker commands, I think that it would be great to do a presentation to the rest of the team about them, given that this is the team's duty to maintain them.

benjben:
I assume we can close this one @istreeter ? I can take care of using your plugin in next release

PR #265, opened 2022-10-25
Warmup improvements

istreeter:


core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Warmup.scala
```diff
@@ -38,30 +40,46 @@ object Warmup {
     if (config.enable) {
       logger.info(s"Starting warm up of $interface:$port.  It is expected to see a few failures during warmup.")
 
-      val cxnSettings = ConnectionPoolSettings(system)
-        .withMaxConnections(config.maxConnections)
-        .withMaxOpenRequests(Integer.highestOneBit(config.maxConnections) * 2) // must exceed maxConnections and must be a power of 2
-        .withMaxRetries(0)
+      def go(counter: Int): Future[Unit] = {
```
pondzix, line 43: If it's about cycles, maybe we should include that somehow in the method's signature? Method's name or arg's name, sth like that :)

pondzix (approved):

PR #271, opened 2022-12-08
PubSub: add integration tests

### LINKED ISSUE #270: PubSub: make it possible to use PubSub emulator
Similarly to AWS with `localstack`, Google provides a [Pub/Sub emulator](https://cloud.google.com/pubsub/docs/emulator) for local tests.

Using it is normally as simple as [running the Docker image](https://cloud.google.com/sdk/docs/downloads-docker) and [setting `PUBSUB_EMULATOR_HOST`](https://cloud.google.com/pubsub/docs/emulator#manually_setting_the_variables).

But [not for Java/Scala](https://cloud.google.com/pubsub/docs/emulator#accessing_environment_variables) : 

```
In all languages except for Java and C#, if you have set PUBSUB_EMULATOR_HOST as described in [Setting environment variables](https://cloud.google.com/pubsub/docs/emulator#env), the Pub/Sub client libraries automatically call the API running in the local instance rather than Pub/Sub.

However, C# and Java client libraries require you to modify your code to use the emulator.
```

We need to add code that modifies the behavior if `PUBSUB_EMULATOR_HOST` is set.

benjben:


pubsub/src/it/scala/com/snowplowanalytics/snowplow/collectors/scalastream/pubsub/GooglePubSubCollectorSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+          )
+          _ <- IO(printBadRows(testName, collectorOutput.bad))
+        } yield {
+          collectorOutput.good.size should beEqualTo(nbGood)
+          collectorOutput.bad.size should beEqualTo(nbBad)
+        }
+      }
+    }
+
+    "shutdown when it receives a SIGTERM" in {
```
istreeter, line 83: Just thinking out loud... I wonder if this should be "shutdown within XX seconds when it receives a SIGTERM".  And then change to `Containers.waitUntilStopped(collector).timeout(10.seconds)`.

pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/GooglePubSubSink.scala
```diff
@@ -100,34 +111,49 @@ object GooglePubSubSink {
...
-    )
+    retrySettings: RetrySettings,
+    customProviders: Option[(TransportChannelProvider, CredentialsProvider)]
+  ): Either[Throwable, Publisher] = {
+    val builder = Publisher
+      .newBuilder(ProjectTopicName.of(projectId, topicName))
+      .setBatchingSettings(batchingSettings)
+      .setRetrySettings(retrySettings)
+      .setHeaderProvider(FixedHeaderProvider.create("User-Agent", UserAgent))
+    val customized = customProviders.fold(builder) {
```
istreeter, line 151: We do this kind of thing a lot in snowplow code, where we take the java builder pattern and then we pretend that the builder is an immutable val.  We force ourselves to use things like `.fold` so we can say we're being functional.   I wonder though if it's better to just be honest that builders are mutable.  And instead do something like this:

```scala
val builder = Publisher
      .newBuilder(ProjectTopicName.of(projectId, topicName))
  // lines skipped for brevity

customProviders.foreach { credentialsProvider =>
  case (channelProvider, credentialsProvider) =>
        builder.setChannelProvider(channelProvider).setCredentialsProvider(credentialsProvider)
}

builder.build()
```

istreeter (approved):
Looks good!

pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/GooglePubSubSink.scala
benjben: > we pretend that the builder is an immutable val

I actually forgot that it is not the case, good point!

PR #277, opened 2023-01-17
WIP: sink-health endpoint should return 503 until proven healthy

istreeter:
This is a work-in-progress PR, so we can test out the implications of starting out unhealthy.


benjben:
Duplicate of https://github.com/snowplow/stream-collector/pull/294

PR #290, opened 2023-02-17
Integration tests for /health endpoint and custom paths

benjben:
Most interesting part is about having only one `localstack` container for all `specs2` `Specification`s, for which we can't define a global `beforeAll` and `afterAll` (unless I missed something).

So I created [this trait](https://github.com/snowplow/stream-collector/blob/feature/test_health/kinesis/src/it/scala/com/snowplowanalytics/snowplow/collectors/scalastream/it/kinesis/containers/Localstack.scala#L30-L34) that tests extend and I [use a `Semaphore`](https://github.com/snowplow/stream-collector/blob/feature/test_health/kinesis/src/it/scala/com/snowplowanalytics/snowplow/collectors/scalastream/it/kinesis/containers/Localstack.scala#L69-L73) to detect if a test is the last one and should stop `localstack`.

### DIFF
--- a/core/src/it/scala/com/snowplowanalytics/snowplow/collectors/scalastream/it/EventGenerator.scala
+++ b/core/src/it/scala/com/snowplowanalytics/snowplow/collectors/scalastream/it/EventGenerator.scala
@@ -14,46 +14,49 @@
  */
 package com.snowplowanalytics.snowplow.collectors.scalastream.it
 
-import scala.concurrent.ExecutionContext
+import cats.effect.IO
 
-import cats.implicits._
-
-import cats.effect.{ContextShift, IO, Resource}
-
-import org.http4s.{Request, Method, Uri}
-import org.http4s.client.Client
-import org.http4s.client.blaze.BlazeClientBuilder
+import org.http4s.{Method, Request, Uri}
 
 object EventGenerator {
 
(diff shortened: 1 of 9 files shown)


spenes (approved):
Looks good to me as it is but another alternative is using [specs2 Resource trait](https://etorreborre.github.io/specs2/guide/5.0.0-RC-05/org.specs2.guide.Contexts.html). However, the problem is that it is introduced in 5.x. We are in 4.x currently. It is up to you to upgrade it.

benjben:
> another alternative is using [specs2 Resource trait](https://etorreborre.github.io/specs2/guide/5.0.0-RC-05/org.specs2.guide.Contexts.html).

As I understand this is quite similar to the existing `BeforeSpec` / `AfterSpec`, as this is just for **one** `Specification`, whereas in my case I want a resource shared by **many** `Specification`s.

spenes:
[In the end of the page](https://etorreborre.github.io/specs2/guide/5.0.0-RC-05/org.specs2.guide.Contexts.html), it is explained how to keep a resource open across several specifications invocation using `resourceKey`.

benjben:
It [seems](https://central.sonatype.com/artifact/org.specs2/specs2-core_2.12/4.19.2) that support for Scala 2.12 and 2.13 stops at `4.19.2`

Issue #276, opened 2023-01-13
Update /sink-health endpoint logic


ianarsenault:
Currently the collector's `/sink-health` endpoint returns with a 200 response immediately upon startup, even if there's an ongoing outage or there are no sinks available with `ResourceNotFoundException` errors. It isn't until after multiple retries, that the endpoint is updated to 503 service unavailable.  

An example I've seen is collectors would start up with `/sink-health` as "healthy" then take 5-10 minutes to become unhealthy after retrying 3 times (with length in between retries depending on backoff). This is not correct from a standpoint of failover, as a health check endpoint should start out unhealthy until proven to be healthy (sinks exist and are available / no outage)

In my testing of using `/sink-health` as a health check for target group I've noticed two areas of concern

1. If traffic spike causes scaling action during an outage -> new collectors that start up `/sink-health` will return 200 until they are updated to be unhealthy after multiple retries (from outage or non-existent resource) - If there's logic based on `/sink-health` to handle failover to set of healthy collectors - this will cause events to be sent to a collector experiencing an outage 
2. During an outage -> Collector services could fall into potential restart loop due to memory or cpu circuit breaks -> causing node’s collector service to restart ->  making it so upon restart `/sink-health` returns healthy/200 and begins taking traffic -> does multiple retries after some time -> updates to 503/unhealthy 

A small snippet of logs show that upon startup there's errors writing but `/sink-health` will return 200 until retries complete
```
NOTE: Picked up JDK_JAVA_OPTIONS: -Dlog4j2.formatMsgNoLookups=true -Xms2g -Xmx2g -Dorg.slf4j.simpleLogger.defaultLogLevel=warn 
--
[pool-1-thread-4] ERROR com.snowplowanalytics.snowplow.collectors.scalastream.sinks.KinesisSink - Writing failed with error:
com.amazonaws.services.kinesis.model.ResourceNotFoundException: Stream collector-raw-stream under account <account_id> not found. (Service: AmazonKinesis; Status Code: 400; Error Code: ResourceNotFoundException; Request ID: c6a451c3-6d25-323f-9a9f-9517f28eea7d; Proxy: null)
[pool-1-thread-4] ERROR com.snowplowanalytics.snowplow.collectors.scalastream.sinks.KinesisSink - Retrying writing batch to Kinesis stream collector-raw-stream in 3000 milliseconds...
--
[...]

PR #294, opened 2023-03-03
Update /sink-health endpoint logic

benjben:
Details in #276

### DIFF
--- a/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
+++ b/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2022 Snowplow Analytics Ltd. All rights reserved.
+ * Copyright (c) 2013-2023 Snowplow Analytics Ltd. All rights reserved.
  *
  * This program is licensed to you under the Apache License Version 2.0,
  * and you may not use this file except in compliance with the Apache License Version 2.0.
@@ -16,31 +16,25 @@ package sinks
 import java.nio.ByteBuffer
 import java.util.concurrent.ScheduledExecutorService
 import java.util.UUID
-import com.amazonaws.auth._
-import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration
-import com.amazonaws.services.kinesis.{AmazonKinesis, AmazonKinesisClientBuilder}
-import com.amazonaws.services.kinesis.model._
-import com.amazonaws.services.sqs.{AmazonSQS, AmazonSQSClientBuilder}
-import com.amazonaws.services.sqs.model.{
-  MessageAttributeValue,
-  QueueDoesNotExistException,
-  SendMessageBatchRequest,
-  SendMessageBatchRequestEntry
-}
-import org.slf4j.LoggerFactory
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ListBuffer
 import scala.concurrent.{ExecutionContextExecutorService, Future}
 import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
+
 import cats.syntax.either._
+
+import com.amazonaws.auth._
+import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration
+import com.amazonaws.services.kinesis.{AmazonKinesis, AmazonKinesisClientBuilder}
+import com.amazonaws.services.kinesis.model._
+import com.amazonaws.services.sqs.{AmazonSQS, AmazonSQSClientBuilder}
+import com.amazonaws.services.sqs.model.{MessageAttributeValue, SendMessageBatchRequest, SendMessageBatchRequestEntry}
+
 import com.snowplowanalytics.snowplow.collectors.scalastream.model._
 import com.snowplowanalytics.snowplow.collectors.scalastream.sinks.KinesisSink.SqsClientAndName
 
-/**
-  * Kinesis Sink for the Scala Stream Collector.
-  */
 class KinesisSink private (
   val maxBytes: Int,
   client: AmazonKinesis,
@@ -54,12 +48,10 @@ class KinesisSink private (
 
   maybeSqs match {
     case Some(sqs) =>
-      log.info(
-        s"SQS buffer for '$streamName' Kinesis sink is set up as: ${sqs.sqsBufferName}."
-      )
+      log.info(s"SQS buffer for Kinesis stream $streamName is defined with name ${sqs.sqsBufferName}")
     case None =>
       log.warn(
-        s"No SQS buffer for surge protection set up (consider setting a SQS Buffer in config.hocon)."
+        s"No SQS buffer for surge protection set up for stream $streamName (consider setting it via the config file)"
       )
   }
 
@@ -77,9 +69,9 @@ class KinesisSink private (
   implicit lazy val ec: ExecutionContextExecutorService =
     concurrent.ExecutionContext.fromExecutorService(executorService)
 
-  // Is the collector detecting an outage downstream
-  @volatile private var outage: Boolean = false
-  override def isHealthy: Boolean       = !outage
+  @volatile private var kinesisHealhy: Boolean = false
+  @volatile private var sqsHealthy: Boolean    = false
+  override def isHealthy: Boolean              = kinesisHealhy || sqsHealthy
 
   override def storeRawEvents(events: List[Array[Byte]], key: String): Unit =
     events.foreach(e => EventStorage.store(e, key))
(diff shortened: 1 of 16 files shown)


kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -142,110 +134,111 @@ class KinesisSink private (
...
-              log.error(s"Sending batch to Kinesis stream $streamName in $nextBackoff milliseconds...")
-            }
+  def handleKinesisError(failedRecords: List[Events], nextBackoff: Long, retriesLeft: Int): Unit =
+    if (retriesLeft > 0) {
+      log.error(
+        s"$retriesLeft retries left. Retrying to write ${failedRecords.size} records to Kinesis stream $streamName in $nextBackoff milliseconds"
+      )
+      scheduleRetryToKinesis(failedRecords, nextBackoff, retriesLeft - 1)
+    } else {
+      kinesisHealhy = false
```
istreeter, line 216: This line surprised me the most.  It is possible that all requests to kinesis were "successful" (returned a success response) but we ran out of retries because we got throttled by kinesis rate limits too many times.

So this changes our definition of unhealthy.  In previous versions, the sink was only marked as unhealthy if kinesis returned error responses.  Whereas now the sink could be marked unhealthy during a traffic spike because of throttling.

I think that might be OK... because we still have the sqs buffer, which sets overall health to healthy.  So I'm just checking this change is deliberate, and you have fully thought through the consequences?

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -410,6 +451,8 @@ object KinesisSink {
             executorService,
             sqsClientAndName
           )
+        ks.checkKinesisHealth()
+        ks.checkSqsHealth()
```
istreeter, line 455: Can you confirm these threads don't interfere with shutting down the app cleanly?

The `KinesisSink` has a `shudown` method, and instinct tells me we should use it to stop the running threads.  Or, run the threads on a executorService, and use the shutdown hook to close the exeuctorService.

But...... whatever works is OK for me.

istreeter (commented):
This is very tidy.  It's much easier to read than the previous implementation of this feature.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -352,14 +349,60 @@ class KinesisSink private (
...
+          } match {
+            case Success("ACTIVE") =>
+              log.info(s"Stream $streamName ACTIVE")
+              kinesisHealhy = true
+            case Success(other) =>
+              log.warn(s"Stream $streamName not ACTIVE but $other")
+              Thread.sleep(1000L)
+            case Failure(err) =>
+              log.error(s"Error while checking status of stream $streamName: ${err.getMessage()}")
+              Thread.sleep(1000L)
```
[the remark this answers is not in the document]
benjben: This is deliberate. The quotas for `describeStream` are per second and thus if we want to maximize the number of successful replies that we get we should be doing the calls once every second (unless I'm missing something).

Let's imagine that we have 100 collector instances starting at once. The first second, 20 instances will get a successful response, and the 80 other ones a quota exceeded exception. The 2nd second, 80 requests will be made again, 20 successful and 60 failures that will be retried the next second, etc. It's impossible to be quicker than this.

On top of that, all the collector instances are configured in the same way. Even if we use a different value, there are good chances that they will end up doing the requests at the same time.

Just as a reminder, this runs in the background and is not blocking, and if events are sent to Kinesis/SQS before the check succeeds, the sink is marked as healthy and the requests are stopped.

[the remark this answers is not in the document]
istreeter: And while all that is happening, any other service (non-collector) running within the account is unable to call `describeStream`, because the collector has stole all the quota.  For example, enrich-kinesis will give an error [on this line](https://github.com/snowplow/enrich/blob/3.7.0/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L90) and then it will exit with an error.

I'm fine with your reasoning, as long as you've thought through how it will impact other services.  I just thought that by making it configurable, we can make the collector less aggressive at using up all the quota, and we might appreciate having that flexibility if we notice odd things happening elsewhere.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
benjben: > This line surprised me the most. It is possible that all requests to kinesis were "successful" (returned a success response) but we ran out of retries because we got throttled by kinesis rate limits too many times.

Yes it's possible!

> In previous versions, the sink was only marked as unhealthy if kinesis returned error responses. Whereas now the sink could be marked unhealthy during a traffic spike because of throttling.

As I understand [there was already no distinction between throughput exceeded errors and internal errors](https://github.com/snowplow/stream-collector/blob/2.8.2/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala#L187-L204). Am I missing something?

> I think that might be OK... because we still have the sqs buffer, which sets overall health to healthy.

I also think that this is how the behavior should be : if we hit throughput exceeded 3 times, we are in a spike, and we want to write to SQS in this case, if set up.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
benjben: Good point, I hadn't thought of that, because these checks happen at startup and should be long done when the collector exits, but we never know, let me update !

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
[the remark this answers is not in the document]
benjben: > For example, enrich-kinesis will give an error [on this line](https://github.com/snowplow/enrich/blob/3.7.0/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L90) and then it will exit with an error.

Good point! I think that we should remove that from Enrich, because we know that our infra configures the streams correctly. I think that we should let Enrich try to write to the streams and fail if it doesn't succeed. I'll create the issue.

>  I just thought that by making it configurable, we can make the collector less aggressive at using up all the quota, and we might appreciate having that flexibility if we notice odd things happening elsewhere.

You're right, I'll make this configurable, it can only be a good thing.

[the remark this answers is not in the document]
benjben: I've just remembered that collector and Enrich won't interfere because collector [uses describeStreamSummary](https://github.com/snowplow/stream-collector/blob/feature/sink_health/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala#L360) and enrich [uses describeStream](https://github.com/snowplow/enrich/blob/3.7.0/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L90) and [they have different limits](https://docs.aws.amazon.com/streams/latest/dev/service-sizes-and-limits.html).

I'll still make it configurable.

kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
istreeter: > As I understand [there was already no distinction between throughput exceeded errors and internal errors](https://github.com/snowplow/stream-collector/blob/2.8.2/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala#L187-L204). Am I missing something?

I don't read it that way, but maybe I'm reading it wrong.  In the old version, if you look for the lines of code which set `outage = true`, I think it never happens if Kinesis always returns a successful result.

But don't worry about it -- this change (if it is a change) only affects collectors with no sqs buffer.  For collectors with a sqs buffer then I'm more confident there is no change compared to the old behaviour.

benjben: I think that it's a good change for collector without SQS buffer.

The variable is now named `kinesisHealthy` (unlike `outage` before), like `/sink-health`, and if writes are throttled, I think that it's not healthy behavior and we should know about it, because the stream might be under provisioned.

istreeter: That's good reasoning, I'm happy with that.

istreeter (approved):

PR #297, opened 2023-03-09
Scan Docker images with Snyk container monitor in deploy.yml 

lmath:
Whenever we release the collector, we run a Snyk scan to check for security vulnerabilites.

Previously, we had a separate Github action for this, `snyk.yml` which just ran `snyk monitor` and did not scan the docker images.

We are moving over to using `snyk container monitor` because we want to
- be able to keep track of vulnerabilities in the docker images we create on release without manually adding them in the Snyk UI
- continue scanning our java/scala jar files

Note that even though we are already creating docker images in `deploy.yml`, they are pushed to the remote registry only, and that is why here we additionally add a step to create local Docker images for the Snyk scan.


lmath:
This is currently a draft because

I expected the vulns that ended up in the dashboard for e.g. `2.8.3-rc9-distroless` to be similar to the current ones for `2.8.2-distroless`, but they are not.

I don't yet know why. I thought maybe the underlying image had changed, so I tried to add an import for `2.8.3-rc9-distroless` via Snyk's docker hub integration to compare, but this failed.

Also, the snyk scan added to `deploy.yml` currently pushes to the old Snyk dashboard

.github/workflows/deploy.yml
```diff
@@ -137,3 +138,18 @@ jobs:
           platforms: linux/amd64,linux/arm64/v8
           tags: ${{ steps.distroless-meta.outputs.tags }}
           push: true
+
+      - name: Build local distroless image, which is needed to run Snyk
+        if: fromJSON(env.RUN_SNYK)
```
lmath, line 143: Otherwise `env.RUN_SNYK` isn't a boolean. More elegant suggestions are quite welcome.

.github/workflows/deploy.yml
```diff
@@ -137,3 +138,18 @@ jobs:
...
+        run: sbt "project ${{ matrix.platform }}Distroless" docker:publishLocal
+      - name: Show local docker images
+        run: docker images
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/docker@master
+        if: fromJSON(env.RUN_SNYK)
+        with:
+          image: "snowplow/scala-stream-collector-${{ matrix.platform }}:${{ github.ref_name }}-distroless"
+          args: "--app-vulns --org=data-cap"
```
lmath, line 152: This still needs to be changed to point to the new prod snyk org

.github/workflows/deploy.yml
istreeter: Maybe you can get this working using [an expanded matrix configuration](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations).  Something like:

```
  matrix:
    platform:
        - kinesis
        - sqs
        - pubsub
        - kafka
        - nsq
        - stdout
        - rabbitmq
    include:
      - platform: kinesis
        run_snyk: true
      - platform: pubsub
        run_snyk: true
```

lmath: Yep, this seems to work nicely. Thanks for the suggestion.

istreeter (approved):
Thanks @lmath!

lmath:
Re: 
> I expected the vulns that ended up in the dashboard for e.g. 2.8.3-rc9-distroless to be similar to the current ones for 2.8.2-distroless, but they are not.

I was comparing them at the end of the day and was simply confused by how one ordering was libraries, docker image and the other was docker image, libraries 🤦‍♀️ 

In the fresh light of a new day we can see that the library vulnerabilities matched and the scan of the newer base image had fewer vulns, as we'd expect 
scan from the existing process:
image
scan via these changes:
image

PR #300, opened 2023-03-10
Ensure integration tests expect tags like x.x.x-distroless

istreeter:


build.sbt
```diff
@@ -60,23 +60,25 @@ lazy val buildInfoSettings = Seq(
 
 // Make package (build) metadata available within source code.
```
benjben, line 61: Could we add in the comment that this is only for integration testing please?

benjben (approved):
Looks great !

PR #319, opened 2023-08-01
Run background check for Kinesis if it is made unhealthy and SQS buff…

benjben:
Here is the testing that I've done:
- Start the collector with both Kinesis stream and SQS buffer
- Send events
- Check that they are sent to Kinesis
- Delete Kinesis stream
- Send events
- Check that they are sent to SQS
- Create Kinesis stream
- Send events
- Check that they are sent to Kinesis
- Delete Kinesis stream
- Send events
- Check that they are sent to SQS
- Create Kinesis stream
- Send events
- Check that they are sent to Kinesis

### DIFF
--- a/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
+++ b/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
@@ -213,16 +213,26 @@ class KinesisSink private (
       )
       scheduleRetryToKinesis(failedRecords, nextBackoff, retriesLeft - 1)
     } else {
-      kinesisHealthy = false
       log.error(s"Maximum number of retries reached for Kinesis stream $streamName for ${failedRecords.size} records")
       maybeSqs match {
         case Some(sqs) =>
           log.error(
             s"SQS buffer ${sqs.sqsBufferName} defined for stream $streamName. Retrying to send the events to SQS"
           )
+          // If Kinesis was already unhealthy, the background check is already running.
+          // It can happen when the collector switches back and forth between Kinesis and SQS.
+          if (kinesisHealthy) {
+            this.synchronized {
+              if (kinesisHealthy) {
+                kinesisHealthy = false
+                checkKinesisHealth()
+              }
+            }
+          }
           writeBatchToSqsWithRetries(failedRecords, sqs, minBackoff, maxRetries)
         case None =>
           log.error(s"No SQS buffer defined for stream $streamName. Retrying to send the events to Kinesis")
+          kinesisHealthy = false
           writeBatchToKinesisWithRetries(failedRecords, maxBackoff, maxRetries)
       }
     }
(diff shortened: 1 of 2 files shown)


kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala
```diff
@@ -213,16 +213,20 @@ class KinesisSink private (
...
       log.error(s"Maximum number of retries reached for Kinesis stream $streamName for ${failedRecords.size} records")
       maybeSqs match {
         case Some(sqs) =>
           log.error(
             s"SQS buffer ${sqs.sqsBufferName} defined for stream $streamName. Retrying to send the events to SQS"
           )
+          // If Kinesis was already unhealthy, the background check is already running.
+          // It can happen when the collector switches back and forth between Kinesis and SQS.
+          if (kinesisHealthy) checkKinesisHealth()
+          kinesisHealthy = false
```
istreeter, line 225: Your code is probably OK.  Almost certainly OK.  Almost certain that nothing will ever go wrong.

But.....

We should remind ourselves that `checkKinesisHealth()` starts a new Runnable to run asynchonrously.  And we use a executor that has >1 thread.  So it's possible that the new task will run on a different thread to this one.

Next, remember that the the operating system is allowed to pause and resume threads whenever it wants to.  So it is technically possible (although incredibly unlikely) that the background thread can set `kinesisHealthy = true` **before** the current thread sets `kinesisHealthy = false`.

I know this is very pedantic, but it's good practice to write the code to show that we're aware about concurrent effects.  In this case the fix is simply this:

```scala
if (kinesisHealthy) {
  kinesisHealthy = false
  checkKinesisHealthy()
}
```

---

There's more.

It is technically possible that two separate threads are both running `handleKinesisError` at the same time as each other.  So it is technically possible (although incredibly unlikely) that two separate threads both start the background thread at the same time as each other.

I think if we were to be really pedantic about this, then the correct implementation is something like:

```scala
if (kinesisHealthy) {
  this.synchronized {
    if (kinesisHealthy) {
      kinesisHealthy = false
      checkKinesisHealthy()
    }
  }
}
```

benjben: Thanks a lot for the detailed comment @istreeter !

I tried to think if something could go wrong. What is not acceptable is that no background check starts whereas it should, and that should never be the case. I thought that even if extremely unlucky and it will likely never happen, it is acceptable that 2 background checks run simultaneously. But you're right that it's better to not let this happen. And I now think that technically there is a possibility that the whole thread pool is busy with background checks, even though **extremely** unlucky. I'll apply your changes.

Murphy's law: _"Anything that can go wrong will go wrong."_

benjben: (smart use of the additional `if (kinesisHealthy) {` at first to not use `synchronized` if not needed)

istreeter (approved):

PR #323, opened 2023-08-07
Http4s - Add post endpoint

spenes:
This PR adds post endpoint. At the moment, it only has basic features to get the request, create the CollectorPayload and sinking the payload. It doesn't include details such as redirection, cookie bounce, do-not-track cookie support etc. These will be included later on.

### DIFF
--- a/http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
+++ b/http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
@@ -1,0 +1,170 @@
+package com.snowplowanalytics.snowplow.collectors.scalastream
+
+import java.util.UUID
+
+import scala.collection.JavaConverters._
+
+import cats.effect.Sync
+import cats.implicits._
+
+import org.http4s.{Request, RequestCookie, Response}
+import org.http4s.Status._
+
+import org.typelevel.ci._
+
+import com.snowplowanalytics.snowplow.CollectorPayload.thrift.model1.CollectorPayload
+
+import com.snowplowanalytics.snowplow.collectors.scalastream.model._
+
+trait Service[F[_]] {
+  def cookie(
+    queryString: Option[String],
+    body: F[Option[String]],
+    path: String,
+    cookie: Option[RequestCookie],
+    userAgent: Option[String],
+    refererUri: Option[String],
+    hostname: F[Option[String]],
+    ip: Option[String],
+    request: Request[F],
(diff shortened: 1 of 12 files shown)


http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -0,0 +1,170 @@
...
+    refererUri: Option[String],
+    hostname: F[Option[String]],
+    ip: Option[String],
+    request: Request[F],
+    pixelExpected: Boolean,
+    doNotTrack: Boolean,
+    contentType: Option[String] = None,
+    spAnonymous: Option[String] = None
+  ): F[Response[F]]
+  def determinePath(vendor: String, version: String): String
```
istreeter, line 35: It seems to me there is an alternative implementation:

- Remove `determinePath` from this trait.
- `cookieRoutes` invokes `collectorService.cookie(...)` using the original path.
- The collector service determines the fixed path from the original path.

As for which one is better.....  we should consider the tests.  Why is `Service` defined as a trait in the fist place?  I believe the answer is so that we can mock it during tests.  A unit test of `CollectorRoutes` using a mocked `Service` is simpler if the `Service` trait definition is minimized to the essentials.

This is a matter of taste though, so feel free to disagree.

istreeter (commented):
Looks good!  Just one idea for your consideration, but I'd be happy to approve it like this.

http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
spenes: `cookie` function is used with [pixel endpoints](https://github.com/snowplow/stream-collector/blob/master/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoute.scala#L107) as well in the akka collector. And pixel tracking doesn't use `determinePath` function. Therefore, it isn't included on `cookie` function in there. I tried to follow same structure. If you think it is okay to use `determinePath` with pixel endpoints as well, I guess we can remove it from `Service` trait.

istreeter: OK let's leave it like it is.  Part of me still thinks we haven't picked the right abstractions for the collector.  But on the other hand we know this way will work, and I shouldn't make this task more complicated than it already is!!!

istreeter (approved):

PR #325, opened 2023-08-08
Configure set-cookie header

spenes:
This PR brings necessary changes to configure set-cookie header.

### DIFF
--- a/http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
+++ b/http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
@@ -2,12 +2,15 @@ package com.snowplowanalytics.snowplow.collectors.scalastream
 
 import java.util.UUID
 
+import scala.concurrent.duration._
 import scala.collection.JavaConverters._
 
-import cats.effect.Sync
+import cats.effect.{Clock, Sync}
 import cats.implicits._
 
-import org.http4s.{Request, RequestCookie, Response}
+import org.http4s._
+import org.http4s.headers._
+import org.http4s.implicits._
 import org.http4s.Status._
 
 import org.typelevel.ci._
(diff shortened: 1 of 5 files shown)


http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
```diff
@@ -27,7 +29,21 @@ object model {
     */
   final case class SplitBatchResult(goodBatches: List[List[Json]], failedBigEvents: List[Json])
 
-  final case class CollectorConfig(
-    paths: Map[String, String]
+  final case class CookieConfig(
+    enabled: Boolean,
+    name: String,
+    expiration: FiniteDuration,
+    domains: Option[List[String]],
```
istreeter, line 36: Is there any significant difference between `None` and `Some(Nil)`?  Did you consider making this a non-optional `List[String]` with a default of `Nil`?

istreeter (commented):
👍 Nice to see an important part of the collector functionality in the http4s project.

Just a couple of questions to think about.

http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -152,6 +166,100 @@ class CollectorService[F[_]: Sync](
...
+        case Some(_) => None
+        case None =>
+          cookieConfig.map { config =>
+            val responseCookie = ResponseCookie(
+              name    = config.name,
+              content = networkUserId,
+              expires = Some(HttpDate.unsafeFromEpochSecond((now + config.expiration).toSeconds)),
+              domain  = cookieDomain(headers, config.domains, config.fallbackDomain),
+              path    = Some("/"),
+              sameSite = config.sameSite.flatMap {
```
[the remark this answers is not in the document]
spenes: Yep, you are right. It should be easy thing to do. I will change that.

http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala
spenes: I don't think there is a difference between `None` and `Some(Nil)`. We can use `List[String]` without wrapping it to `option` I guess. I will make necessary changes.

istreeter (approved):

PR #326, opened 2023-08-09
Load config

benjben:
- [x] Make config path optional


http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Config.scala
```diff
@@ -0,0 +1,293 @@
...
+
+  def parse[F[_]: Sync](
+    args: List[String]
+  ): F[Config] = {
+    val appConfig = Opts.option[Path]("config", "Path to HOCON configuration", "c", "config.hocon")
+    val parser    = Command[Path]("collector", "Snowplow application that collects tracking events")(appConfig)
+    for {
+      path <- parser.parse(args) match {
+        case Left(help) =>
+          Sync[F].raiseError(new IllegalArgumentException(s"can't read CLI arguments. $help"))
```
istreeter, line 31: This is OK for production environments... but I don't think it's the best user experience to get an exception and a stack trace if the command line is invalid.  For the lake loader I used `CommandIOApp` instead of `IOApp` and I was pleased with how it worked out.  It's documented [here](https://ben.kirw.in/decline/effect.html).  If the user supplies the wrong command line config then it handles showing a helpful error message and a non-zero exit code.

project/Dependencies.scala
```diff
@@ -48,14 +48,17 @@ object Dependencies {
     val akkaHttpMetrics  = "1.7.1"
     val badRows          = "2.1.1"
     val log4cats         = "2.6.0"
+    val http4s           = "0.23.23"
+    val blaze            = "0.23.15"
+    val http4sNetty      = "0.5.9"
+    val decline          = "2.4.1"
+    val circe            = "0.13.0"
```
istreeter, line 55: Why not circe 0.14.1 and circeConfig 0.10.0?

http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Config.scala
```diff
@@ -0,0 +1,293 @@
...
+import cats.effect.Sync
+
+import com.monovore.decline.{Command, Opts}
+
+object Config {
+
+  def parse[F[_]: Sync](
+    args: List[String]
+  ): F[Config] = {
+    val appConfig = Opts.option[Path]("config", "Path to HOCON configuration", "c", "config.hocon")
```
istreeter, line 26: For the existing collector, the `--config` param is optional.  So that you can configure the collector using system properties as an alternative to providing a file.

e.g.

```
java -Dcollector.streams.good=good -jar collector.jar
```

But here I think you've made it a required parameter.

http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Config.scala
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  private def parseConfigFile[F[_]: Sync](path: Path): F[Config] =
+    Either.catchNonFatal(Files.readAllLines(path).asScala.mkString("\n")) match {
+      case Right(raw) =>
+        val config = ConfigFactory
+          .parseString(raw)
+          .resolve()
+          .withFallback(ConfigFactory.load("application.conf"))
+          .withFallback(ConfigFactory.load("reference.conf"))
```
istreeter, line 46: For all our other snowplow applications, we accept config parameters with the following order of precedence:

1. System properties set with the `-D` command line option
2. The provided config file (which is optional)
3. `application.conf`
4. `reference.conf`

Here you have allowed for 2-4 but you haven't allowed for system properties.

The way to do this is to sandwich the layers of configuration like the old collector does [here](https://github.com/snowplow/stream-collector/blob/2.9.1/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Collector.scala#L82).

istreeter (approved):

http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Config.scala
benjben: I was only focusing on knowing the error to solve the issue, not on the style. I'll make it look nicer like you did.

http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/Config.scala
benjben: I didn't think of 1, I'll update :+1:

project/Dependencies.scala
benjben:
```
[info]   +-com.snowplowanalytics:snowplow-badrows_2.12:2.1.1
[info]   | +-com.snowplowanalytics:iglu-scala-client_2.12:1.1.1
[info]   | +-com.snowplowanalytics:snowplow-scala-analytics-sdk_2.12:2.1.0 [S]
[info]   | | +-com.snowplowanalytics:iglu-core-circe_2.12:1.0.0 [S]
[info]   | | | +-com.snowplowanalytics:iglu-core_2.12:1.0.0 [S]
[info]   | | | +-io.circe:circe-core_2.12:0.13.0 (evicted by: 0.14.1)
```
I just need to bump bad rows to 2.2.1 :+1:

http4s/src/main/scala/com.snowplowanalytics.snowplow.collector.core/Service.scala
```diff
@@ -39,26 +39,24 @@ trait Service[F[_]] {
...
 }
 
-class CollectorService[F[_]: Sync](
-  config: CollectorConfig,
-  sinks: CollectorSinks[F],
-  appName: String,
-  appVersion: String
-) extends Service[F] {
+class Service[F[_]: Sync, SinkConfig](
+  config: Config[SinkConfig],
```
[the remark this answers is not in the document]
benjben: Great!

spenes (commented):
Looks great! It might be good to have some config tests as well. Other than tests, it is good to go I think.

PR #327, opened 2023-08-10
Pubsub sink

pondzix:
* full implementation of PubSub sink: producing events asynchronously + background pubsub health check
* `core` part of integration tests based on CE2 has been copied to the new `http4s` module and ported to CE3.


http4s/src/main/scala/com.snowplowanalytics.snowplow.collector.core/ConfigParser.scala
```diff
@@ -67,7 +67,7 @@ object ConfigParser {
   }
 
   private def loadAll(config: TypesafeConfig): TypesafeConfig =
-    namespaced(ConfigFactory.load(namespaced(config.withFallback(namespaced(ConfigFactory.load())))))
+    namespaced(config.withFallback(namespaced(ConfigFactory.load())))
```
benjben, line 70: See https://github.com/snowplow/stream-collector/pull/334#discussion_r1298101653

pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/PubSubHealthCheck.scala
```diff
@@ -0,0 +1,74 @@
...
+      _ <- repeatInBackgroundUntilHealthy(isHealthyState, sinkConfig, healthCheckTask)
+    } yield ()
+
+  private def repeatInBackgroundUntilHealthy[F[_]: Async](
+    isHealthyState: Ref[F, Boolean],
+    sinkConfig: PubSubSinkConfig,
+    healthCheckTask: F[Unit]
+  ): Resource[F, Unit] = {
+    val checkThenSleep = healthCheckTask *> Async[F].sleep(sinkConfig.startupCheckInterval)
+    checkThenSleep.untilM_(isHealthyState.get).background.void
```
benjben, line 37: :ok_hand:

pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/PubSubSink.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  override def storeRawEvents(events: List[Array[Byte]], key: String): F[Unit] =
+    produceBatch(events).start.void
+
+  override def isHealthy: F[Boolean] = isHealthyState.get
+
+  private def produceBatch(events: List[Array[Byte]]): F[Unit] =
+    events.parTraverse_ { event =>
+      produceSingleEvent(event) *> isHealthyState.set(true)
+    }
```
benjben, line 53:
```suggestion
    events.parTraverse_ { event =>
      produceSingleEvent(event)
    }  *> isHealthyState.set(true)
```
What about setting the sink healthy only after the whole batch has been succesfully inserted ?

pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/PubSubSink.scala
```diff
@@ -0,0 +1,136 @@
...
+class PubSubSink[F[_]: Async: Parallel: Logger] private (
+  override val maxBytes: Int,
+  isHealthyState: Ref[F, Boolean],
+  producer: PubsubProducer[F, Array[Byte]],
+  retryInterval: FiniteDuration,
+  topicName: String
+) extends Sink[F] {
+
+  override def storeRawEvents(events: List[Array[Byte]], key: String): F[Unit] =
+    produceBatch(events).start.void
```
benjben, line 46: If the tracker doesn't batch the events, `events` will always be of size 1. But nevertheless this seems fine to me to start a `Fiber` for each single event.

pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/PubSubSink.scala
```diff
@@ -0,0 +1,136 @@
...
+  private def produceSingleEvent(event: Array[Byte]): F[Model.MessageId] =
+    producer
+      .produce(event)
+      .retryingOnAllErrors(
+        policy  = RetryPolicies.constantDelay(retryInterval),
+        onError = (error, _) => handlePublishError(error)
+      )
+
+  private def handlePublishError(error: Throwable): F[Unit] =
+    isHealthyState.set(false) *> Logger[F].error(createErrorMessage(error))
```
benjben, line 64: Related to the discussion about logging policy, we'll need to decide if we should be logging errors for individual events or if we need some batching.

benjben (approved):
Looks great @pondzix !

pubsub/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/ConfigSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+  "Config parser" should {
+    "be able to parse extended pubsub config" in {
+      parse("/config.pubsub.extended.hocon") must beRight(ConfigSpec.expectedConfig)
+    }
+    "be able to parse minimal pubsub config" in {
+      parse("/config.pubsub.minimal.hocon") must beRight(ConfigSpec.expectedConfig)
+    }
+  }
+
+  private def parse(resource: String): Either[ExitCode, Config[PubSubSinkConfig]] = {
```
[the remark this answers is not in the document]
pondzix: I think I would keep it here for now. When we add more sinks and we indeed keep repeating this piece of code, we can extract to common shared module.

pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/PubSubSink.scala
```diff
@@ -0,0 +1,136 @@
...
+    error match {
+      case apiEx: ApiException =>
+        val retryable = if (apiEx.isRetryable) "retryable" else "non-retryable"
+        s"Publishing message to $topicName failed with code ${apiEx.getStatusCode} and $retryable error: ${apiEx.getMessage}"
+      case throwable => s"Publishing message to $topicName failed with error: ${throwable.getMessage}"
+    }
+}
+
+object PubSubSink {
+  private val UserAgent = s"snowplow/stream-collector-${BuildInfo.version}"
```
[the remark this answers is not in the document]
pondzix: Shouldn't that be just `BuildInfo.dockerAlias`? It looks like it already contains version

pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/PubSubHealthCheck.scala
```diff
@@ -0,0 +1,74 @@
...
+        Logger[F].info(s"Topic $topicName exists") *> isHealthyState.set(true)
+      case Right(false) =>
+        Logger[F].error(s"Topic $topicName doesn't exist")
+      case Left(err) =>
+        Logger[F].error(s"Error while checking if topic $topicName exists: ${err.getCause}")
+    }
+
+  private def createTopicAdminClient[F[_]: Sync](): Resource[F, TopicAdminClient] = {
+    val builder = TopicAdminSettings.newBuilder().setProvidersForEmulator().build()
+    Resource.make(Sync[F].delay(TopicAdminClient.create(builder)))(client => Sync[F].delay(client.close()))
```
istreeter, line 57: Do you know if `client.close()` is blocking or non-blocking?

pubsub/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/PubSubSink.scala
istreeter: Good point.  This only becomes an issue if there is an outage and if it lasts for a long time.  Ideally we should find a nice pattern in the code for de-duping log messages upon errors.  And we should start using it everywhere when we've decided on that pattern.

No need to hold up this current PR though.

PR #328, opened 2023-08-10
Add get and head endpoints

spenes:


http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -45,6 +47,10 @@ class CollectorService[F[_]: Sync](
   appVersion: String
 ) extends Service[F] {
 
+  // Contains an invisible pixel to return for `/i` requests.
+  val pixelBase64 = "R0lGODlhAQABAPAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw=="
+  val pixel       = Stream.emit[F, String](pixelBase64).through(text.base64.decode)
```
istreeter, line 52: I think here you base64-decode the pixel for every request we receive, and the result is the same each time.  Would be nice to arrange the code so we have the pixel as an array of bytes when the collector first starts up.

For example, could use `java.util.Base64` to decode it when you first start up.  Or could store the raw pixel in the resources directory, and load it when you first start up.

http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala
```diff
@@ -153,6 +177,12 @@ class CollectorService[F[_]: Sync](
       }
     }
 
+  /** If the pixel is requested, this attaches cache control headers to the response to prevent any caching. */
+  def cacheControl(pixelExpected: Boolean): Option[`Cache-Control`] =
+    if (pixelExpected)
+      Some(`Cache-Control`(CacheDirective.`no-cache`(), CacheDirective.`no-store`, CacheDirective.`must-revalidate`))
```
istreeter, line 183: This line is very important!  I had forgotten the collector needs to do this, or else trackers will not send all requests to the collector.

Do we have a test somewhere (new unit test or old integration test) which checks these are set?

spenes: I couldn't find any mention of that in the integration tests but there are some [checks](https://github.com/snowplow/stream-collector/blob/master/core/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorServiceSpec.scala#L217) for that in the unit tests. I've added specific unit test for that in the new application's test suite as well.

istreeter (approved):

PR #329, opened 2023-08-11
Add pixel endpoint

spenes:


http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoutes.scala
```diff
@@ -19,51 +18,41 @@ class CollectorRoutes[F[_]: Sync](collectorService: Service[F]) extends Http4sDs
...
             doNotTrack    = false,
-            contentType   = None,
-            spAnonymous   = spAnonymous
+            contentType   = None
           )
       }
+
+    case req @ (GET | HEAD) -> Root / ("ice.png" | "i") =>
+      collectorService.cookie(
+        body          = Sync[F].delay(None),
```
istreeter, line 47: Probably doesn't make much difference, but this could be `.pure` instead of `.delay`.

http4s/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorRoutes.scala
```diff
@@ -19,51 +18,41 @@ class CollectorRoutes[F[_]: Sync](collectorService: Service[F]) extends Http4sDs
 
   private val cookieRoutes = HttpRoutes.of[F] {
     case req @ (POST | GET | HEAD) -> Root / vendor / version =>
```
istreeter, line 20: I like how with these new changes the routes class is very clean and visually easy to read.

What do you think about splitting out the `/vendor/version` endpoint into separate POST and GET -- would this look even more clean?

```scala
  case req @ POST -> Root / vendor / version =>
    ???

  case req @ (GET | HEAD) -> Root / vendor / version =>
    ???

  case case req @ (GET | HEAD) -> Root / ("ice.png" | "i") =>
    ???
```

istreeter (approved):
Just a couple of ideas, but low-priority.

PR #334, opened 2023-08-17
Fix config loading

benjben:


spenes (approved):
I think this is okay but it would be good to get @istreeter's opinion as well since original design was his and there might be reason to use `ConfigFactory.load` in there.

http4s/src/main/scala/com.snowplowanalytics.snowplow.collector.core/ConfigParser.scala
```diff
@@ -67,7 +67,7 @@ object ConfigParser {
   }
 
   private def loadAll(config: TypesafeConfig): TypesafeConfig =
-    namespaced(ConfigFactory.load(namespaced(config.withFallback(namespaced(ConfigFactory.load())))))
+    namespaced(config.withFallback(namespaced(ConfigFactory.load())))
```
istreeter, line 70: If you do it this way, then config precedence goes in this order:

1. The config file
2. System properties passed on the command line (e.g. `-Dcollector.port=8888`)
3. Application defaults

It is more common to have this kind of precedence:

1. System properties
2. The config file
3. Application defaults.

All of our other applications have the extra `ConfigFactory.load` so I would prefer to keep it if we can find a way.

benjben: The reason for this change is that we observed that the config file was not overriding the default parameters. Do you know why ?

istreeter: #336 seems to fix it.

We might need to look more closely at `namespaced` to check it's reliable and to check whether it is overall helpful for us.  But for now I suggest merge in #336 because that's roughly what we do in other apps.  And we can revisit it later when we have more time.

PR #335, opened 2023-08-18
HTTPS support

spenes:


peel (approved):
It may be a subject for subsequent PR (a separate ticket?), but it'd be great to verify that the behaviour actually works with the shape of certificates we expect instead of manual testing. We used to have a clunky way of doing that, but we could possibly have an integration test that generates `jks` and runs the app to check that TLS works. 

Edit: Potentially in CI we could even rely on valid ACME certs for this and avoid all the untrusted cert issues.

spenes:
Yep, that would be good. I will create separate ticket for that.

PR #338, opened 2023-08-23
Add CE3 based Kinesis sink

pondzix:


istreeter (commented):
This is looking good.

However, the old collector had quite a intricate interplay between the kinesis and sqs queues.  Here you have chosen to base your sink from the Enrich implementation, not the old collector implementation.  That might be a good decision, but I don't know yet!  We might find it's difficult to patch-in the SQS buffer into this implementation.

Time will tell....

pondzix:
Closing in favour of https://github.com/snowplow/stream-collector/pull/339

PR #339, opened 2023-08-24
Old Kinesis sink with cats effect

pondzix:


kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/KinesisCollector.scala
```diff
@@ -14,57 +14,41 @@
...
-      good <- KinesisSink.createAndInitialize(
-        kc.maxBytes,
-        kc,
-        bufferConf,
-        goodStream,
-        sqsGood,
-        es
+  override def mkSinks(config: Config.Streams[KinesisSinkConfig]): Resource[IO, Sinks[IO]] = {
+    val threadPoolExecutor = buildExecutorService(config.sink)
+    for {
```
istreeter, line 31: How about

```
for {
  stpe <- Resource.make(Sync[F].delay(buildExecutorService(config.sink))(stpe => Sync[F].delay(stpe.shutdown()))
  good <- ....
```

pondzix: Isn't executor closed [here](https://github.com/snowplow/stream-collector/blob/85723b9606bb56a439281a3032f6cee740351c30/kinesis/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KinesisSink.scala#L349) as a part of sink shutdown?

istreeter: Ah yes.  That's fine then feel to ignore my comment.

If we were designing this from scratch then I would say better to shutdown the service very close in the code to where it is created.

istreeter (approved):

PR #343, opened 2023-08-25
Don't use Java properties to pass configuration in  kinesis integrati…

pondzix:
…on tests


kinesis/src/it/scala/com/snowplowanalytics/snowplow/collectors/scalastream/it/core/DoNotTrackCookieSpec.scala
```diff
@@ -62,20 +64,19 @@ class DoNotTrackCookieSpec extends Specification with Localstack with CatsEffect
           headers must haveSize(2)
           expected.forall(cookie => headers.exists(_.contains(cookie))) must beTrue
         }
-      }
-    }
-
-    "track events that have a cookie whose name and value match doNotTrackCookie config if disabled" in {
+      }.unsafeRunSync()
+    }.pendingUntilFixed("Remove when 'do not track cookie' feature is implemented")
```
benjben, line 68: Have you added this info in do not track ticket?

benjben (approved):
LGTM!

kinesis/src/it/scala/com/snowplowanalytics/snowplow/collectors/scalastream/it/core/DoNotTrackCookieSpec.scala
pondzix: Good point, I'll do that

PR #345, opened 2023-09-06
Wrap sqs sink with effects

spenes:
Follows the same way [in this PR](https://github.com/snowplow/stream-collector/pull/339)

### DIFF
--- a/sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
+++ b/sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
@@ -12,6 +12,11 @@
  */
 package com.snowplowanalytics.snowplow.collectors.scalastream.sinks
 
+import cats.effect.{Resource, Sync}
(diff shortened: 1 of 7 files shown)


sqs/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/SqsSink.scala
```diff
@@ -307,7 +332,7 @@ object SqsSink {
   }
 
   /** Create an aws credentials provider through env variables and iam. */
-  private def getProvider(awsConfig: AWSConfig): Either[Throwable, AWSCredentialsProvider] = {
+  private def getProvider(awsConfig: SqsSinkConfig.AWSConfig): Either[Throwable, AWSCredentialsProvider] = {
```
istreeter, line 335: I know this is unrelated to your PR, but I really dislike this `getProvider` method.  At Snowplow we always use the default chain, and I cannot think of any situation when we would use any other method.  I would vote to remove this unhelpful functionality as part of the new collector release.

istreeter (approved):
Looks good, this should work.

PR #349, opened 2023-09-13
Wrap nsq sink with cats effects (close #348)

### LINKED ISSUE #348: Wrap nsq sink with cats effects

colmsnowplow:
Tested manually as follows:

Deploy nsq locally:

```
docker run --rm --name nsqd -p 4150:4150 -p 4151:4151 \
    --network nsqNet \
    nsqio/nsq /nsqd \
    --broadcast-address=localhost
```

Build docker asset locally:

```
sbt 'project nsq' docker:publishLocal
```

Deploy it:

```
docker run --rm \
  --name collector \
  --network nsqNet \
  -v $PWD/config.hocon:/snowplow/config.hocon \
  -p 8080:8080 \
  snowplow/scala-stream-collector-nsq:3.0.0-rc1-24-7c3c9eac --config /snowplow/config.hocon 
//replace with what publishLocal prints for version
```

Send test data:

```
curl 'http://localhost:8080/i?e=pv' --globoff
```

Check it got to nsq:

```
 curl localhost:4151/stats
```

### DIFF
--- a/nsq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/NsqSink.scala
+++ b/nsq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/NsqSink.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2022 Snowplow Analytics Ltd.
+ * Copyright (c) 2013-2023 Snowplow Analytics Ltd.
  * All rights reserved.
  *
  * This program is licensed to you under the Apache License Version 2.0,
@@ -19,17 +19,31 @@
 package com.snowplowanalytics.snowplow.collectors.scalastream
 package sinks
 
+import java.util.concurrent.TimeoutException
+
 import scala.collection.JavaConverters._
 
+import cats.effect.{Resource, Sync}
+import cats.implicits._
+
 import com.snowplowanalytics.client.nsq.NSQProducer
-import com.snowplowanalytics.snowplow.collectors.scalastream.model._
+import com.snowplowanalytics.snowplow.collector.core.{Sink}
+import com.snowplowanalytics.client.nsq.exceptions.NSQException
 
 /**
   * NSQ Sink for the Scala Stream Collector
   * @param nsqConfig Configuration for Nsq
   * @param topicName Nsq topic name
   */
-class NsqSink(val maxBytes: Int, nsqConfig: Nsq, topicName: String) extends Sink {
+class NsqSink[F[_]: Sync] private (
+  val maxBytes: Int,
+  nsqConfig: NsqSinkConfig,
+  topicName: String
+) extends Sink[F] {
+
+  @volatile private var healthStatus = true
+
+  override def isHealthy: F[Boolean] = Sync[F].pure(healthStatus)
 
   private val producer = new NSQProducer().addAddress(nsqConfig.host, nsqConfig.port).start()
 
@@ -38,9 +52,27 @@ class NsqSink(val maxBytes: Int, nsqConfig: Nsq, topicName: String) extends Sink
     * @param events The list of events to send
     * @param key The partition key (unused)
     */
-  override def storeRawEvents(events: List[Array[Byte]], key: String): Unit =
-    producer.produceMulti(topicName, events.asJava)
+  override def storeRawEvents(events: List[Array[Byte]], key: String): F[Unit] =
+    Sync[F].blocking(producer.produceMulti(topicName, events.asJava)).onError {
+      case _: NSQException | _: TimeoutException =>
+        Sync[F].delay(healthStatus = false)
+    } *> Sync[F].delay(healthStatus = true)
 
-  override def shutdown(): Unit =
+  def shutdown(): Unit =
     producer.shutdown()
+}
+
+object NsqSink {
+
+  def create[F[_]: Sync](
+    nsqConfig: NsqSinkConfig,
+    topicName: String
+  ): Resource[F, NsqSink[F]] =
+    Resource.make(
+      Sync[F].delay(
+        // MaxBytes is never used but is required by the sink interface definition,
+        // So just pass any int val in.
+        new NsqSink(0, nsqConfig, topicName)
+      )
+    )(sink => Sync[F].delay(sink.shutdown()))
 }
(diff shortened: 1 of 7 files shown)


nsq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/NsqSink.scala
```diff
@@ -19,28 +19,80 @@
...
 
+  def produceData(topicName: String, events: List[Array[Byte]]): Unit =
+    try {
+      producer.produceMulti(topicName, events.asJava)
+    } catch {
+      case e @ (_: NSQException | _: TimeoutException) => {
+        healthStatus = false
+        throw e
+      }
+    }
```
benjben, line 64:
```suggestion
def produceData(topicName: String, events: List[Array[Byte]]): F[Unit] =
    Sync[F].blocking(producer.produceMulti(topicName, events.asJava))
      .onError { case _: NSQException | _: TimeoutException =>
        Sync[F].delay(healthStatus = false)
      } *> Sync[F].delay(healthStatus = true)
```
Requires to `import cats.implicits._`

Health status was never made healthy again.

benjben: But without retries a failure will crash the app so there isn't much point in setting the health status to `false` it will never be queried

nsq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/NsqSink.scala
```diff
@@ -19,28 +19,80 @@
...
+    executorService: ScheduledExecutorService
+  ): Resource[F, NsqSink[F]] = {
+    val acquire = Sync[F].delay(
+      new NsqSink(maxBytes, nsqConfig, topicName, executorService)
+    )
+
+    val release = (sink: NsqSink[F]) => (Sync[F].delay(sink.shutdown()))
+
+    Resource.make(acquire)(release)
+  }
```
benjben, line 97:
```suggestion
 ): Resource[F, NsqSink[F]] =
    Resource.make {
      Sync[F].delay(new NsqSink(maxBytes, nsqConfig, topicName, executorService))
    }{ sink =>
      Sync[F].delay(sink.shutdown())
    }
```

nsq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/NsqSinkConfig.scala
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2013-2022 Snowplow Analytics Ltd.
```
benjben, line 2:
```suggestion
 * Copyright (c) 2013-2023 Snowplow Analytics Ltd.
```

benjben (commented):
I know that you tested manually but given that Mini will still live for a long time it'd be great to add an integration test. I can help you with that if you need. It would just be a combination of [what we have in Enrich](https://github.com/snowplow/enrich/tree/master/modules/nsq/src/it/scala/com/snowplowanalytics/snowplow/enrich/nsq/test) and what we have for the collector in an other sink ([e.g. Kinesis](https://github.com/snowplow/stream-collector/blob/master/kinesis/src/it/scala/com/snowplowanalytics/snowplow/collectors/scalastream/it/kinesis/KinesisCollectorSpec.scala#L53)). Not a strong requirement though

nsq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/NsqSinkConfig.scala
benjben: :point_up:

nsq/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/NsqConfigSpec.scala
```diff
@@ -18,8 +18,108 @@
...
+    cookieBounce = Config.CookieBounce(
+      enabled                 = false,
+      name                    = "n3pc",
+      fallbackNetworkUserId   = "00000000-0000-4000-A000-000000000000",
+      forwardedProtocolHeader = None
+    ),
+    redirectMacro = Config.RedirectMacro(
+      enabled     = false,
+      placeholder = None
+    ),
```
benjben, line 94: Instead of redefining everything you can probably reuse [what is in `TestUtils`](https://github.com/snowplow/stream-collector/blob/feature/http4s/http4s/src/test/scala/com.snowplowanalytics.snowplow.collector.core/TestUtils.scala#L27) and copy it : `TestUtils.testConfig.copy()` (e.g. [here](https://github.com/snowplow/stream-collector/blob/feature/http4s/http4s/src/test/scala/com.snowplowanalytics.snowplow.collector.core/ServiceSpec.scala#L372-L374))

benjben (approved):
:+1:

nsq/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/NsqSink.scala
benjben: Thanks for testing! Sorry this was my mistake. This is very bad behavior, because in case of an issue with NSQ, the collector drops the events forever and never crashes. It should either retry the insert forever until it blows up or crash after the first failure (and restart). But given that this is just for Mini and the old collector was already behaving like this, let's keep as is.

colmsnowplow: No apology needed - how you expected it to be is how it should be, I absolutely agree!

nsq/src/test/scala/com.snowplowanalytics.snowplow.collectors.scalastream/NsqConfigSpec.scala
colmsnowplow: I tried, but couldn't get the type conversion to work initially - we don't do this for any of the other config tests, so in the interest of getting it merged I'll leave as is if that's ok.

PR #352, opened 2023-09-20
Make maxConnections and idleTimeout configurable

peel:
Previusly, idleTimeout has been hardcoded, maxConnections hasn't been configured.
Now, these parameters are set within `networking` section and used throughout http4s backends.


http4s/src/main/scala/com.snowplowanalytics.snowplow.collector.core/HttpServer.scala
```diff
@@ -33,20 +32,24 @@ object HttpServer {
     app: HttpApp[F],
     interface: String,
     port: Int,
-    secure: Boolean
+    secure: Boolean,
+    maxConnections: Int,
+    idleTimeout: FiniteDuration
```
benjben, line 37: What about passing `networking` ? Building the HTTP server probably will likely always require all the parameters of `Networking`

http4s/src/main/resources/reference.conf
```diff
@@ -77,6 +77,11 @@
     port = 443
   }
 
+  networking {
+    maxConnections = 1024
```
benjben, line 81: In prod we use `8192` by default

benjben (approved):

http4s/src/main/scala/com.snowplowanalytics.snowplow.collector.core/HttpServer.scala
peel: netty does not require maxConnections, but I guess it can very well be `networking` instead of the two separate params 👍

http4s/src/main/resources/reference.conf
peel: Yea, I wanted to keep it the same as the defaults in akka until we properly calculate what makes the most sense for our processing patterns. Good spot regardless.

PR #353, opened 2023-09-25
Http4s Kafka Sink

spenes:


kafka/src/it/scala/com/snowplowanalytics/snowplow/collectors/scalastream/it/kafka/Containers.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.testcontainers.containers.wait.strategy.Wait
+import org.testcontainers.containers.{BindMode, Network, GenericContainer => JGenericContainer}
+
+object Containers {
+
+  val zookeeperContainerName = "zookeeper"
+  val zookeeperPort = 2181
+  val brokerContainerName = "broker"
+  val brokerExternalPort = 9092
+  val brokerInternalPort = 29092
```
pondzix, line 17: Shouldn't that be...external? It doesn't really matter, but isn't it usually like internal port (here 9092) is mapped to some host's port (here 29092)? And couldn't we make it random instead of using fixed host port container? 

It might be also the case I misunderstood something as I don't know details about kafka <-> zookeeper communication, then feel free to ignore this comment!

spenes: Actually, internal means internal to docker network in here. I know it isn't very good naming but I couldn't find better.
Kafka Collector Docker instance communicating via 29092 port of broker through docker network. 9092 port is used by test to fetch the records from Kafka topic.

> couldn't we make it random instead of using fixed host port container?

We need to use fixed port because we need to specify ports to connect to broker in `KAFKA_ADVERTISED_LISTENERS`.

pondzix (approved):

PR #402, opened 2023-12-13
collector-kafka: authenticate with Event Hubs using OAuth2

spenes:


kafka/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/AzureAuthenticationCallbackHandler.scala
```diff
@@ -0,0 +1,71 @@
...
+  val credentials: MSICredentials = new MSICredentials()
+
+  var sbUri: String = ""
+
+  override def configure(
+    configs: util.Map[String, _],
+    saslMechanism: String,
+    jaasConfigEntries: util.List[AppConfigurationEntry]
+  ): Unit = {
+    val bootstrapServer = configs.get(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG).toString.replaceAll("\\[|\\]", "")
```
istreeter, line 39: Out of curiosity, please can you share an example of what this bootstrap server config looks like?  I'm curious about which part of it needs replacing/removing.

[The Kafka documentation](https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#bootstrap-servers) says bootstrap servers can be a comma separated list of multiple servers.  Should we also split by comma and take the first host?

kafka/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/AzureAuthenticationCallbackHandler.scala
```diff
@@ -0,0 +1,71 @@
...
+    val uri             = URI.create("https://" + bootstrapServer)
+    this.sbUri = uri.getScheme + "://" + uri.getHost
+  }
+
+  override def handle(callbacks: Array[Callback]): Unit =
+    callbacks.foreach {
+      case callback: OAuthBearerTokenCallback =>
+        val token = getOAuthBearerToken()
```
istreeter, line 47: Is the callbacks array ever `!= 1` in length?  If yes, I would suggest to move `getOAuthBearerToken` above `callbacks.foreach`.  But if it is always array of length 1 then what you've done is good.

istreeter (commented):
Have you checked how often we end up calling `credentials.getCredentials()`?  E.g. by adding some temporary logging during development.

Presumably (hopefully) something somewhere is caching the token, so the app doesn't fetch a new token for every event.  If it was me I would want to see proof of that while working on the feature.

Overall this looks great though 👍

kafka/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/AzureAuthenticationCallbackHandler.scala
spenes: > Out of curiosity, please can you share an example of what this bootstrap server config looks like? I'm curious about which part of it needs replacing/removing.

It looks like `host1:port1,host2:port2`. I don't think it has any `[]` but I've added `replaceAll` part just in case since they had that in [the example](https://github.com/Azure/azure-event-hubs-for-kafka/blob/master/tutorials/oauth/java/managedidentity/consumer/src/main/java/CustomAuthenticateCallbackHandler.java#L41).

> [The Kafka documentation](https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#bootstrap-servers) says bootstrap servers can be a comma separated list of multiple servers. Should we also split by comma and take the first host?

Yes, it should be done like that. Good catch! I will update it.

kafka/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/AzureAuthenticationCallbackHandler.scala
spenes: Hmm, I think there should be at most one `OAuthBearerTokenCallback` callback but I will move it above just in case.

spenes: Actually, if we move it above, `getOAuthBearerToken` will be called even if there is no `OAuthBearerTokenCallback` therefore it would be better to leave it as it is.

istreeter (approved):

kafka/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/AzureAuthenticationCallbackHandler.scala
```diff
@@ -0,0 +1,85 @@
...
+        .get(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG)
+        .toString
+        .replaceAll("\\[|\\]", "")
+        .split(",")
+        .toList
+        .headOption match {
+        case Some(s) => s
+        case None    => throw new Exception("Empty bootstrap servers list")
+      }
+    val uri = URI.create("https://" + bootstrapServer)
```
ianarsenault, line 54: I am curious why we set the auth uri against the eventhubs bootstrap server from the producer conf? The managed identity / service account is at the active directory level so it might need to be updated use https://login.microsoftonline.com/ or possibly https://management.core.windows.net/.  There's a [powershell](https://github.com/logcorner/Getting-an-access-token-inside-an-aks-pod-using-Azure-Managed-Identities-with-Workload-Identity-Fede/blob/main/my-script.ps1) example of what I believe we're trying to achieve.

For example, we deploy a [service account](https://github.com/snowplow-devops/helm-charts/blob/main/charts/cloudserviceaccount/templates/serviceaccount.yaml#L13) in AKS which adds the following ENV variables on the collector pod

```
AZURE_AUTHORITY_HOST: https://login.microsoftonline.com/
AZURE_CLIENT_ID: aabb44cc69f-e118-4636-b21b-f43598fdkj39fh33jdjfk   
AZURE_FEDERATED_TOKEN_FILE: /var/run/secrets/azure/tokens/azure-identity-token
AZURE_TENANT_ID: 333333-b20a-4bda-ade0-3423432443
```

The AZURE_AUTHORITY_HOST is `https://login.microsoftonline.com/` so I would've expected this to be what we would auth against for a token.

The `AZURE_CLIENT_ID` and `AZURE_FEDERATED_TOKEN_FILE` are both important here as `AZURE_CLIENT_ID` is the managed identity ID and `AZURE_FEDERATED_TOKEN_FILE` contains  token for the service account 

```
{
  "aud": [
    "api://AzureADTokenExchange"
  ],
  "exp": 1703283977,
  "iat": 1703280377,
  "iss": "https://eastus.oic.prod-aks.azure.com/d21c78e8-a114-11ee-8c90-0242ac120002
/d8519c0c-a114-11ee-8c90-0242ac120002/",
  "kubernetes.io": {
    "namespace": "collector-namespace",
    "pod": {
      "name": "my-collector-pod-5f8f495dc7-rnk5h",
      "uid": "ee25c8d2-a114-11ee-8c90-0242ac120002"
    },
    "serviceaccount": {
      "name": "collector-serice-account",
      "uid": "fbb17f6e-a114-11ee-8c90-0242ac120002"
    }
  },
  "nbf": 1703280377,
  "sub": "system:serviceaccount:collector-namespace:collector-service-account"
}
```

With these values using azure cli we can generate a token like 

```
az login --federated-token "$(cat $AZURE_FEDERATED_TOKEN_FILE)" --service-principal -u $AZURE_CLIENT_ID -t $AZURE_TENANT_ID
```

and then run the needed az commands to interact with whatever azure resources have been granted access to the managed identity.

spenes: We are using `uri` with [MSICredentials.getToken](https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-authentication/src/main/java/com/microsoft/azure/credentials/MSICredentials.java) method in `getOAuthBearerToken`. According to my understanding, it is more like a resource id than auth uri. We are using bootstrap server address as resource name as [documented here](https://learn.microsoft.com/en-us/azure/event-hubs/authenticate-application).

However, the problem we are facing at the moment I think is using `MSICredentials`. Since we want to deploy Collector in AKS, we have to use Workload Identity auth however it isn't possible with `MSICredentials`. If we use [DefaultAzureCredential](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/identity/azure-identity/src/main/java/com/azure/identity/DefaultAzureCredential.java) to get token, it should be possible to use Workload Identity auth as well.

kafka/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KafkaSink.scala
```diff
@@ -88,6 +90,7 @@ object KafkaSink {
       props.setProperty("linger.ms", bufferConfig.timeLimit.toString)
       props.setProperty("key.serializer", "org.apache.kafka.common.serialization.StringSerializer")
       props.setProperty("value.serializer", "org.apache.kafka.common.serialization.ByteArraySerializer")
+      props.setProperty("sasl.login.callback.handler.class", authCallbackClass)
```
istreeter, line 93: By setting to the property here, I think you are forcing **all** kafka users (including open-source users) to use Azure authentication.  Don't we anticipate some users want to integrate with non-azure kafka?

The previous implementation set the handler class in the `application.conf` file, so there was a standard way for anyone to override the setting with something different.  Is there a reason that configuration approach won't work anymore?

spenes: It can still be overwritten actually since producer properties from config is set later on. 

I made it like this because this property is a bit special one. It have to be a class name from the application. Adding this property to `application.conf` looks a bit unsafe because in that case, there won't be any direct reference to callback handlers. Someone later on might remove those classes and application would compile without any problem but this feature would be broken. However, with current implementation, we directly reference the callback handler classes and if they removed from the project, the project wouldn't even compile. Of course, you can say that we might add some test cases to make sure they aren't removed but this implementation looks better to me. 

I don't have strong opinion about this one, I can change it the way you described as well if you want.

istreeter: > It can still be overwritten actually since producer properties from config is set later on.

Ah, my mistake... I mis-read the code earlier.  You are right, it can be overridden.

I'm happy then.

istreeter (approved):

PR #404, opened 2023-12-28
Add statsd metrics reporting

pondzix:
* Following metrics are aggregated:

```
active_requests
headers_time
requests_count
requests_latency
abnormal_count
abnormal_latency
```

Using [datadog4s](https://avast.github.io/datadog4s/userguide.html#http4s) library providing http4s metrics middleware. It’s slightly different than the ones provided by [Akka](https://github.com/RustedBones/akka-http-metrics/blob/a70120f2a3f894115cebcd5e1cd1449a068cf108/datadog/src/main/scala/fr/davit/akka/http/metrics/datadog/DatadogSettings.scala#L26) HTTP in old collector: 

```
requests_active
requests_bytes
requests_count
responses_bytes
responses_count
responses_duration
```

* Using empty `tags` as default to be consistent with other Snowplow apps.


core/src/main/resources/reference.conf
```diff
@@ -70,6 +70,9 @@
         port = 8125
         period = 10 seconds
         prefix = snowplow.collector
+        tags = {
```
istreeter, line 73: In most other apps we leave the default tags empty, e.g. [here](https://github.com/snowplow-incubator/snowplow-lake-loader/blob/0.1.4/modules/core/src/main/resources/reference.conf#L33).

istreeter (approved):

core/src/main/resources/reference.conf
pondzix: I used nonempty default, because that's what we do for [old](https://github.com/snowplow/stream-collector/blob/fd41ec04665a8dfd68dc7d7e610d49556a35e96d/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/model.scala#L181) collector?

istreeter: True... but we never yet used those metrics internally.  Whereas now there is a plan to start using them.  So now is our chance to get the tags right.

If we forward these metrics to cloudwatch, then each tag becomes a cloudwatch Dimension.  And for cloudwatch you must specify every Dimension when querying a meric.  So by adding a default tag here, it affects how the metrics get queried later.  By leaving tags empty, we can be consistent across all apps when it comes to querying the metrics.

peel (approved):

pondzix:
Merged to `release/3.0.0` and then little push force to update `Prepare for 3.0.0 release` commit with statsd stuff.

Commit on `release/3.0.0` -> https://github.com/snowplow/stream-collector/commit/4ca903bfc3cae7381251247e779a530ecd148329

PR #409, opened 2024-01-16
Add an option to send HSTS header (fix #408)

### LINKED ISSUE #408: Add an option to send an HSTS header with every response
The [HSTS header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) can prevent man-in-the-middle attacks.

When the Collector is behind something like Nginx, the simplest approach is to configure the HSTS header to be added by Nginx.

However, often the Collector is directly behind a cloud load balancer that does not have the option to send this header. So it is useful to be able to send the header from the Collector itself, so that it’s passed through by the load balancer.

stanch:


spenes (approved):
Looks good 👍

PR #410, opened 2024-01-24
Release 3.1.0

peel:


colmsnowplow (approved):
LGTM - I think this is one of the repos where the version is automagically inferred but in case it isn't I don't see a version bump anywhere.

peel:
@colmsnowplow yea, it is [this smart](https://github.com/snowplow/stream-collector/blob/release%2F3.1.0/README.md?plain=1#L46) ;)

PR #411, opened 2024-02-15
Cross compile to scala 2.12

pondzix:
This is implemented, because we need both Collector and Enrich in Micro. It turns out it's easier to ad cross compile to 2.12 in Collector, than upgrade Enrich to 2.13. This should be temporary solution, we need to have everything running on 2.13 eventually.


colmsnowplow (approved):
LGTM

PR #414, opened 2024-02-26
Log cats-effect warning at debug level

spenes:
Implementation is taken from [here](https://github.com/snowplow/enrich/commit/c4ff36ec0e60c7bf4415026d367a721e237915f4).


peel (approved):

peel:
@spenes can you add a bit of a context into a commit body? I think it's useful to keep it close to the code change and here we're dealing with a particular decision that we've taken and an edge case that we know the exists.

spenes:
Sure @peel. Is following enough or should we need anything more ?

```
We need these silenced for now as we investigate other parts of the application and reasons for low resource usage in both collector 3.x and 2.x.
```

PR #415, opened 2024-02-27
Add QA framework job for builds

peel:
QA framework[1] is an end-to-end testing suite for Snowplow services. It runs a desired configuration for pipeline components in a real-life cloud deployment.

This change adds a trigger for automatically running QA framework for tag builds. Why all tag builds? These tags are added for when we decide to evaluate an asset in a cloud environment. Final rc builds don't differ much from stable tags. Therefore, testing the stable asset should be a formality and should not be a blocking step.

Why don't we retag stable asset after testing? This option has been tested. The change is relatively simple in the script - skopeo[2] can retag images without pulling them. However, as an image needs to be accessible for it to be deployed it needs a temporary tag (including a commit ref ideally). This however extends the build and as stable release is usually a retag of the final rc, the benefit is minimal. We may however, revisit this in the future.

Why don't we run it for all PRs? Well, it takes ages to run these things. It's simple enough to create an rc if you need it.

1 - https://github.com/snowplow-devops/qa-framework 
2 - https://github.com/containers/skopeo

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -150,3 +150,28 @@ jobs:
           command: monitor
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+  qa_framework:
+    runs-on: ubuntu-latest
+    needs: deploy_to_docker
+    steps:
+      - uses: actions/checkout@v2
+      - uses: coursier/cache-action@v3
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Get current version
+        id: ver
+        run: |
+          export PROJECT_VERSION=$(sbt "project core" version -Dsbt.log.noformat=true | perl -ne 'print "$1\n" if /info.*(\d+\.\d+\.\d+[^\r\n]*)/' | tail -n 1 | tr -d '\n')
+          echo "project_version=$PROJECT_VERSION" >> $GITHUB_OUTPUT
+      - name: Execute QA Framework tests
+        uses: aurelien-baudet/workflow-dispatch@v2
+        with:
+          workflow: test-component
+          repo: snowplow-devops/qa-framework
+          token: ${{ secrets.GLOBAL_QA_FRAMEWORK_PAT }}
+          ref: "refs/heads/main"
+          inputs: '{ "test_directory": "collector", "qa_collector_version": "${{ steps.ver.outputs.project_version }}" }'
+          wait-for-completion-timeout: 30m # workflow is much slower in Azure; can take up to 20-something minutes
+          display-workflow-run-url: true
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -173,5 +173,5 @@ jobs:
           token: ${{ secrets.GLOBAL_QA_FRAMEWORK_PAT }}
           ref: "refs/heads/main"
           inputs: '{ "test_directory": "collector", "qa_collector_version": "${{ steps.ver.outputs.project_version }}" }'
-          wait-for-completion-timeout: 30m # workflow is much slower in Azure; can take up to 20-something minutes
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -150,3 +150,28 @@ jobs:
...
+          echo "project_version=$PROJECT_VERSION" >> $GITHUB_OUTPUT
+      - name: Execute QA Framework tests
+        uses: aurelien-baudet/workflow-dispatch@v2
+        with:
+          workflow: test-component
+          repo: snowplow-devops/qa-framework
+          token: ${{ secrets.GLOBAL_QA_FRAMEWORK_PAT }}
+          ref: "refs/heads/main"
+          inputs: '{ "test_directory": "collector", "qa_collector_version": "${{ steps.ver.outputs.project_version }}" }'
+          wait-for-completion-timeout: 30m # workflow is much slower in Azure; can take up to 20-something minutes
```
jbeemster, line 176: Would probably suggest 60m as that is the default we use with `go test` under the hood to account for very slow executions.

colmsnowplow:
👋 Just a question about the description:

> Why all tag builds? These tags are added for when we decide to evaluate an asset in a cloud environment. Final rc builds don't differ much from stable tags. Therefore, testing the stable asset should be a formality and should not be a blocking step.

Does this mean that this PR tests all tags but is does not block PRs/fail if the tests fail? Or that it does something different for rc builds vs stable builds? Or something else?

.github/workflows/deploy.yml
```diff
@@ -150,3 +150,28 @@ jobs:
...
+          echo "project_version=$PROJECT_VERSION" >> $GITHUB_OUTPUT
+      - name: Execute QA Framework tests
+        uses: aurelien-baudet/workflow-dispatch@v2
+        with:
+          workflow: test-component
+          repo: snowplow-devops/qa-framework
+          token: ${{ secrets.GLOBAL_QA_FRAMEWORK_PAT }}
+          ref: "refs/heads/main"
+          inputs: '{ "test_directory": "collector", "qa_collector_version": "${{ steps.ver.outputs.project_version }}" }'
+          wait-for-completion-timeout: 60m # workflow is much slower in Azure; can take up to 20-something minutes
```
colmsnowplow, line 176: Am I right in this assumption:

- It waits 60m for the test to finish (in the other repo), and on timeout it fails(?)
- If the workflow in the other repo fails, this one fails
- If the workflow in the other repo succeeds, this one succeeds

peel: Yes

peel:
> Does this mean that this PR tests all tags but is does not block PRs/fail if the tests fail? Or that it does something different for rc builds vs stable builds? Or something else?

So, this PR will trigger QA framework run for all tags that are created - both RC and stable. For both it will fail the build, but it runs as a last step. So it is possible that for a stable build it fails but the asset is available. Therefore, the rationale for running both RC and stable in the same way. 
Alternatively we could have a different path for stable build:
- tag `1.2.3-golden`
- run qa framework
- sync `1.2.3-golden` to `1.2.3`

colmsnowplow:
OK thanks, I think it follow. So:

> So it is possible that for a stable build it fails but the asset is available.

Can we avoid that? If we desire a same workflow for both, how about this:

- tag `1.2.3` or `1.2.3-something` gets pushed
- we build and deploy `1.2.3-test` or `1.2.3-something-test`
- we run the tests
- If it passes, we push a docker tag (with the same image) which matches the original value (`1.2.3` or `1.2.3-something`)
- (Optionally), we delete the `-test` asset

peel:
>  Can we avoid that? If we desire the same workflow for both, how about this:
Yep, needs a skopeo action to sync two different image tags and that's it. But my point is we should probably wait until we're happy with this action until we make it a blocking one. Therefore that's why I have not pushed it in the PR. But I'm happy to revive it in this or another PR.

colmsnowplow (approved):
Forgot to leave the review earlier

PR #417, opened 2024-03-22
Performance improvements

peel:
#### [Add timeout for body parsing](https://github.com/snowplow/stream-collector/commit/6e08cb969d443a98d45b138e2183d7340184cf86) 
For a long-running connections (for example ones coming from a load-balancer), we used to wait for body text stream to complete and therefore hanged the request infinitely. This caused performance issues. Now, we detect if there was no activity for a given body text for the period of time and short-circuit the processing early. We extend timeout in testing to accommodate for virtual environment performance.

#### [Add debug logging and timeout configurations](https://github.com/snowplow/stream-collector/commit/af8f52dac5f3cc7634231860f63138a40441a4fd) 
Previously it was only possible to enable trace logs for requests and responses. These can contain PII and should not be enabled. After this change it is possible to explicitly configure which parts of request we want to log. For now, body is left on or off, but we should provide a better mechanism to filter out PIIs on demand.

#### [Allow setting size limit on line and header length](https://github.com/snowplow/stream-collector/pull/417/commits/3f77728cfaa39f769cfb3aa6accbcedb8b5e382a) 
 
There was a slight difference in behavior between series/3.x and series/2.x where the old one's default values on lengths were higher. Those requests were marked as invalid. Now, the value is configurable.


kafka/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/sinks/KafkaSink.scala
```diff
@@ -40,7 +40,7 @@ class KafkaSink[F[_]: Sync](
     * @param events The list of events to send
     * @param key The partition key to use
     */
-  override def storeRawEvents(events: List[Array[Byte]], key: String): F[Unit] = Sync[F].delay {
+  override def storeRawEvents(events: List[Array[Byte]], key: String): F[Unit] = Sync[F].blocking {
```
istreeter, line 43: This line stood out to me as not ideal.  If we use `Sync[F].blocking` here then every single Event needs to be shifted to the blocking thread pool.  A task then runs on that blocking thread pool, which executes `kafkaProducer.send` (line 46).  After that is finished, processing shifts back onto the first thread pool again where we send the response back to the client.  That's a lot of thread shifting for a lot of events.

I understand that this is a workaround because `kafkaProducer.send` is blocking under some circumstances.  But I was really hoping we could find a way to make that non-blocking, instead of shifting blocking calls to a different thread pool.

Sorry for the negative opinion!!  If this is the only to fix it then of course I'm ok with that.

peel: I've got an alternative approach that uses fs2-kafka's Producer (not a full fs2 stream) that does a two-step operations on it that also seems to be working. I decided to put this one here, but I'll remove it and add another PR with just the fs2-kafka's producer.

peel: @istreeter I've reorganised it a bit, removed the `blocking` commit and organised everything by adding an additional fs2-kafka commit into the PR. I'm happy to split it up into separate PRs.

peel:
For transparency, I've added snowman tests into the PR and included Kafka sink PR (#418) so everything gets run through it: https://github.com/snowplow-devops/snowman/actions/runs/8457939954
Merging into develop and rebasing into release branch

PR #418, opened 2024-03-26
Prevent Kafka sink from blocking

peel:
Previously, we used native Kafka sink implementation through kafka-clients. This caused an issue when actual `send` operation becomes blocking as the record write buffer fills and an actual send operation is performed. We're currently successfully using fs2-kafka in enrich with Enrich without major issues. Rewriting the Sink to use a two-effect pattern makes no sense and therefore fs2-kafka is used to prevent blocking problems.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,24 +1,80 @@
 {
   "nodes": {
-    "devenv": {
+    "cachix": {
       "inputs": {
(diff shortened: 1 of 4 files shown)


istreeter (approved):
Nice solution :)

PR #423, opened 2024-04-17
Allow custom headers to be returned as part of pre-flight requests

miike:
Certain services (e.g., Datadog) have a requirement for adding non CORS-safelisted headers to network requests in order to record tracing information for later reporting.

Currently requests which included non-safelisted headers error out when performing the pre-flight OPTIONS request. This adds the ability to specify a custom list (in addition to the default headers) that are returned as part of the `Access-Control-Allow-Headers` response.

Context: https://discourse.snowplow.io/t/access-to-xmlhttprequest-has-been-blocked-by-cors-policy-request-header-field-traceparent-is-not-allowed-by-access-control-allow-headers-in-preflight-response/9922/3

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -52,6 +52,7 @@
 
   cors {
     accessControlMaxAge = 60 minutes 
+    allowedCorsHeaders = ["X-Example"]
   }
 
   streams {
--- a/core/src/main/scala/com.snowplowanalytics.snowplow.collector.core/Service.scala
+++ b/core/src/main/scala/com.snowplowanalytics.snowplow.collector.core/Service.scala
@@ -17,6 +17,7 @@ import org.apache.commons.codec.binary.Base64
 import scala.concurrent.duration._
 import scala.jdk.CollectionConverters._
 
+import cats.data.NonEmptyList
 import cats.effect.{Clock, Sync}
 import cats.implicits._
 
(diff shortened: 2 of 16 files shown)


core/src/main/scala/com.snowplowanalytics.snowplow.collector.core/Service.scala
```diff
@@ -139,11 +140,13 @@ class Service[F[_]: Sync](
   }
 
   override def preflightResponse(req: Request[F]): F[Response[F]] = Sync[F].pure {
+    val allowedHeaders = (List("Content-Type", "SP-Anonymous") ++ config.cors.allowedCorsHeaders).map(CIString(_))
+    val corsHeaders    = NonEmptyList.fromListUnsafe(allowedHeaders)
```
istreeter, line 144: Think we can remove the "unsafe" method.  Which is harmless in this case, but stands out as worrisome when you see it in collector code.

```scala
val corsHeaders = NonEmptyList("Content-Type", "SP-Anonymous" :: config.cors.allowedCorsHeaders).map(CIString(_))
```

core/src/main/resources/reference.conf
```diff
@@ -52,6 +52,7 @@
 
   cors {
     accessControlMaxAge = 60 minutes 
+    allowedCorsHeaders = ["X-Example"]
```
istreeter, line 55: This file provides defaults for the app.  So by adding "X-Example" here, it means all pipelines everywhere will respond with "X-Example" in the list of allowed headers.

Guessing that's probably not what you want?

core/src/main/scala/com.snowplowanalytics.snowplow.collector.core/Service.scala
```diff
@@ -139,11 +140,13 @@ class Service[F[_]: Sync](
   }
 
   override def preflightResponse(req: Request[F]): F[Response[F]] = Sync[F].pure {
+    val allowedHeaders = (List("Content-Type", "SP-Anonymous") ++ config.cors.allowedCorsHeaders).map(CIString(_))
```
stanch, line 143: What about just making `Content-Type` and `SP-Anonymous` a part of the `allowedCorsHeaders` setting? That would avoid rebuilding this list on every request. And actually `allowedCorsHeaders` is a misnomer if it does not include _all_ allowed headers.

miike: Do you mean as part of the config? My only concern would be if people remove this from a default config.

PR #429, opened 2024-09-03
Remove body read timeout feature

peel:
The goal of the feature was to prevent long body reads in GCP, this however does not prevent the slow incoming connection handling at the framework level. Therefore, as this adds unnecessary complexity with possible negative performance impact, the feature is removed.


pondzix (approved):

core/src/main/scala/com.snowplowanalytics.snowplow.collector.core/Run.scala
```diff
@@ -98,7 +98,6 @@ object Run {
           config.enableDefaultRedirect,
           config.rootResponse.enabled,
           config.crossDomain.enabled,
-          config.networking.responseHeaderTimeout,
```
pondzix, line 101: Now I noticed - we never passed `bodyReadTimeout`? Is it intentional?

peel: Yea. In the end there was no difference between the two. Removing makes much more sense.

PR #433, opened 2024-11-19
Kinesis allow larger batches when sqs buffer is enabled

istreeter:


istreeter:
Implemented by https://github.com/snowplow/stream-collector/commit/9288115b62e4cc232de2e48af29da4d3430a5088

PR #435, opened 2024-11-22
Release/3.2.1

peel:


CHANGELOG.md
```diff
@@ -0,0 +1,622 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
```
colmsnowplow, line 8: Should it have this [Unreleased] bit? Is there an automation that fills it out or something?

colmsnowplow (approved):
LGTM!

I did review the code but ofc not the best placed to spot issues, assuming we are confident in our previous reviews of the implementation.

CHANGELOG.md
peel: According to keepachangelog the header should stay there: https://keepachangelog.com/en/1.1.0/#effort
For now there is no automation. We can get it done separately. The goal for now is to keep changes in readable sections so it's easy to find functional the differences between versions.

colmsnowplow: Fair enough - I don't really like it and would argue that it's just noise on top of the Changelog, which doesn't seem to serve the purpose stated in that doc, because we never merge unreleased changes. IMO the same thing can be achieved by making a new section when we make a new develop branch.

But that's probably just a personal preference, I'm not bothered and wouldn't fight you on it. Certainly wouldn't block a release because of it! Just an opinion for consideration :)

PR #437, opened 2025-01-07
Release/3.3.0

AlexBenny:


CHANGELOG.md
```diff
@@ -13,6 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 ### Security
 
+## [3.3.0] - 2024-12-18
+### Changed
+- Kafka sink use a dedicated thread for potentially blocking send [snowplow/stream-collector-private#3]
+
+### Security
+- Add limit for payload allowed sizes [snowplow/stream-collector-private#1]
+- Security and license updates [snowplow/stream-collector-private#1]
```
stanch, line 22: We probably don’t need these links?

AlexBenny: I'd like to keep the minimum changes in the content as it will be hard to manage both origins if content is different. If not strictly required I would keep them there.

istreeter (approved):

PR #438, opened 2025-02-14
Allow JSON logging options 

### LINKED ISSUE #281: Json log format
In my case SnowpplowCollector run in kubernetes cluster. I heed to easy see logs.

I see options:
- Integration with LogStash
- Json log format

rob-ellison-jet:
Closes: #281

This introduces alternative logging to the SimpleLogger with logback and logstash.

This should support both simple log format and JSON log format.

There is also an option to enable async logging although the queue size is fixed at 500.

I also added an option to set log levels per class for disabling kinesis and pubsub warnings.

I've been struggling to run tests locally as i'm having issues with ryuk running with Podman.

I think perhaps to improve this it would be better to set logger options in the config rather than in the java options.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,3 @@ target
 src/main/resources/application.conf
 
 .bsp
-
-.bloop
-.metals
-.vscode
-
-metals.sbt
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,9 @@ target
 src/main/resources/application.conf
 
 .bsp
+
+.bloop
+.metals
+.vscode
+
+metals.sbt
--- a/core/src/main/resources/logback.xml
+++ b/core/src/main/resources/logback.xml
@@ -1,0 +1,28 @@
+<configuration>
+
+    <!-- Console Appender with dynamic log format -->
+    <appender name="consoleAppender" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
+            <pattern>%level  %logger{36} - %msg%n</pattern> <!-- Simple logger style -->
+        </encoder>
+    </appender>
+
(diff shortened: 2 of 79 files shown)


.gitignore
```diff
@@ -5,3 +5,9 @@ target
 src/main/resources/application.conf
 
 .bsp
+
+.bloop
+.metals
+.vscode
+
+metals.sbt
```
rob-ellison-jet, line 13: Sorry, didn't mean to commit these changes

core/src/main/resources/logback.xml
```diff
@@ -0,0 +1,28 @@
...
+    </appender>
+
+    <!-- JSON Log Appender -->
+    <appender name="jsonConsoleAppender" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>
+    </appender>
+
+    <!-- Async Appender -->
+    <appender name="asyncAppender" class="ch.qos.logback.classic.AsyncAppender">
+        <queueSize>500</queueSize>
```
rob-ellison-jet, line 17: I may be useful to make this value customisable. Potentially a logging section of the config might allow for greater flexibility

core/src/main/resources/logback.xml
```diff
@@ -0,0 +1,28 @@
+<configuration>
+
+    <!-- Console Appender with dynamic log format -->
+    <appender name="consoleAppender" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
+            <pattern>%level  %logger{36} - %msg%n</pattern> <!-- Simple logger style -->
```
rob-ellison-jet, line 6: Not too much thought was put into this pattern. It could probably be done in a better structured format

